PR #1, opened 2017-03-08
Add rake task support

jordanlewis:
This just delegates to the Postgres rake task implementations. This
works fine except for db:structure:dump/load, which delegate to the
`pg_dump` binary. Those raise exceptions for now.

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb/database_tasks.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/database_tasks.rb
@@ -1,0 +1,19 @@
+require "active_record/base"
+
+module ActiveRecord
+  module ConnectionAdapters
+    module CockroachDB
+      class DatabaseTasks < ActiveRecord::Tasks::PostgreSQLDatabaseTasks
+        def structure_dump(filename, extra_flags=nil)
+          raise "db:structure:dump is unimplemented. See https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/2"
+        end
+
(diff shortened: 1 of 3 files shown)


jordanlewis:
PTAL, updated after https://github.com/cockroachdb/cockroach/pull/14006 landed.

lib/activerecord-cockroachdb-adapter.rb
```diff
@@ -0,0 +1,12 @@
+if defined?(Rails)
+  module ActiveRecord
+    module ConnectionAdapters
+      class CockroachDBRailtie < ::Rails::Railtie
+        rake_tasks do
+          # Load rake tasks.
```
tamird: where is this comment cribbed from? I'd just as soon remove it.

lib/active_record/connection_adapters/cockroachdb/database_tasks.rb
```diff
@@ -0,0 +1,19 @@
+require "active_record/base"
+
+module ActiveRecord
+  module ConnectionAdapters
+    module CockroachDB
+      class DatabaseTasks < ActiveRecord::Tasks::PostgreSQLDatabaseTasks
+        def structure_dump(filename, extra_flags=nil)
```
tamird: since you don't care about these arguments, you could write `structure_dump(*)` instead. same applies below.

jordanlewis: We'll care about them pretty soon so I'll leave them.

PR #7, opened 2017-05-14
Fix compatibility issues with Rails 5.1

RickCSong:
In Rails 5.1, #extract_schema_qualified_name is redefined in the
PostgreSQLAdapter.  Because CockroachDBAdapter defines it slightly
differently, this causes compatibility issues between the two.  This
change ensures that the namespaces are no longer colliding.

This also restores the #index_name_length and #table_alias_length methods which
were changed in ActiveRecord 5.1 to use `SHOW max_identifier_length` which does not
exist in CockroachDB.  We hardcode this to 63 (as it was previously in
ActiveRecord 5.0).

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb_adapter.rb
+++ b/lib/active_record/connection_adapters/cockroachdb_adapter.rb
@@ -25,6 +25,13 @@ end
 
 class ActiveRecord::ConnectionAdapters::CockroachDBAdapter < ActiveRecord::ConnectionAdapters::PostgreSQLAdapter
   ADAPTER_NAME = "CockroachDB".freeze
+
+  # Note that in the migration from ActiveRecord 5.0 to 5.1, the
+  # `extract_schema_qualified_name` method was aliased in the PostgreSQLAdapter.
+  # To ensure backward compatibility with both <5.1 and 5.1, we rename it here
+  # to use the same original `Utils` module.
+  Utils = ActiveRecord::ConnectionAdapters::PostgreSQL::Utils
+
   def indexes(table_name, name = nil) # :nodoc:
     # The PostgreSQL adapter uses a correlated subquery in the following query,
     # which CockroachDB does not yet support. That portion of the query fetches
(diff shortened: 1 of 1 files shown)


tamird:
Hi Rick! 👋

lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -117,7 +120,8 @@ def column_definitions(table_name)
       end_sql
   end
 
-  private def extract_schema_qualified_name(*args)
-    ActiveRecord::ConnectionAdapters::PostgreSQL::Utils.extract_schema_qualified_name(*args)
+  def table_alias_length
+    63
```
RickCSong: :+1: Sounds great.  I'll add a comment for this :)

RickCSong:
Updated with comments @tamird !  Thanks for feedback.

lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -117,7 +124,17 @@ def column_definitions(table_name)
       end_sql
   end
 
-  private def extract_schema_qualified_name(*args)
-    ActiveRecord::ConnectionAdapters::PostgreSQL::Utils.extract_schema_qualified_name(*args)
+  # This is hardcoded to 63 (as previously was in ActiveRecord 5.0) to aid in
+  # migration from PostgreSQL to Cockroachdb. In practice, this limitation
+  # is arbitrary since CockroachDB supports index name lengths and table alias
+  # lengths far greater than this value.  For the time being though, we match
```
tamird: two spaces after this period

lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -117,7 +124,17 @@ def column_definitions(table_name)
...
+  # the original behavior for PostgreSQL to simplify migrations.
+  #
+  # Note that in the migration to ActiveRecord 5.1, this was changed in
+  # PostgreSQLAdapter to use `SHOW max_identifier_length` (which does not
+  # exist in CockroachDB).  Therefore, we have to redefine this here.
```
tamird: ditto

jordanlewis:
LGTM once Tamir's comments are resolved. Thanks!

RickCSong:
Done!  Thanks

PR #10, opened 2017-06-13
Don't use deprecated pgconn

jordanlewis:
And actually depend on pg.


benesch:
I think the original revision omitted the pin, but Jordan pushed a new
commit with a pin and the description rotted. ¯\_(ツ)_/¯

jordanlewis:
Yeah, forgot to update PR description. `pg` has weird versioning semantics.

Issue #12, opened 2017-10-19
Uuid for unique ids in Rails 5.x fails due to direct inheritance with PostgreSQL adapter


mcansky:
Rails 5.x has first class support for uuid as unique ID for objects. PostgreSQL 9.4+ has a direct support of that through a call to gen_random_uuid() in the create query :

```
CREATE TABLE "users" ("id" uuid DEFAULT gen_random_uuid() NOT NULL PRIMARY KEY, "email" character varying, [...]
```

Unfortunately that function doesn't exist in the case of cockroachdb. Instead, as pointed out in the documentation (https://www.cockroachlabs.com/docs/v1.1/uuid.html#create-a-table-with-auto-generated-unique-row-ids) uuid_v4()::UUID() should be used.

## a quick work around

As described in the postgresql adapter code (https://github.com/rails/rails/blob/8f037a5ab929f929fe0897edc9ecadf901c32e56/activerecord/lib/active_record/connection_adapters/postgresql/schema_definitions.rb#L7-L45) it's possible to pass a custom function name to handle uuid generation.

So, to get through with the create the migration should look like this :

```
create_table :users, id: false do
  t.primary_key :id, :uuid, default: "uuid_v4()::UUID"
  [...]
end
```

## permanent fix

I'd guess one needs to add something similar to https://github.com/rails/rails/blob/8f037a5ab929f929fe0897edc9ecadf901c32e56/activerecord/lib/active_record/connection_adapters/postgresql/schema_definitions.rb#L46-L52 for this adapter. I might have a try if someone can validate this thought.

tamird:
Looks like gen_random_uuid is just an alias of uuid_v4, perhaps with
stronger "randomness". Maybe the right thing is to implement that function
in CRDB.

jordanlewis:
+1 to @tamird's suggestion - we'll implement that builtin in CockroachDB.

mcansky:
@jordanlewis , @tamird : ok, I am split though, check my PR and close it if you think it's better to keep this adapter as close as possible to the postgresql one.

benesch:
Agreed that we should add the builtin—unless there are cryptographic guarantees that we don't provide?—but that won't make it in until 1.2, no? Patching the adapter would make things work today.

mcansky:
@benesch : patching the adapter would make it work out of the box faster yes; if someone can point me to some documentation or give me a hand in #13 it could be ready soon

[...]

PR #13, opened 2017-10-19
WIP Uuid function fix

mcansky:
Related : https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/12

An attempt to set the uuid function name properly for AR to use when generating the SQL query; unfortunately I am not familiar enough to attach those bits in the adapter to get it to work fine.
It's probably not missing much, any direction into where to look would be appreciated.

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb/schema_definitions.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/schema_definitions.rb
@@ -37,6 +37,10 @@ module ActiveRecord
           super
         end
       end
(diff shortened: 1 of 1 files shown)


mcansky:
closing by lack of interest

jordanlewis:
Sorry I let this drop - see discussion in #12. Again thanks for pitching in and it's my bad for not taking a look at this earlier. The good news is that we'll get a fix into CockroachDB in the next patch release.

PR #16, opened 2017-11-14
schema.rb: fix detection of default int values

flavio:
Some of postgresql's default regexps used by activerecord are not going to work with cockroachdb, leaving the default values out of the `schema.rb` file.

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb_adapter.rb
+++ b/lib/active_record/connection_adapters/cockroachdb_adapter.rb
@@ -145,4 +145,29 @@ class ActiveRecord::ConnectionAdapters::CockroachDBAdapter < ActiveRecord::Conne
     63
   end
   alias index_name_length table_alias_length
+
+
+  private
+
+    # Extracts the value from a PostgreSQL column default definition.
+    # Some of postgresql's default regexps are not going to work with cockroachdb,
+    # leaving the default values out of the schema.rb file.
+    def extract_value_from_default(oid, default) # :nodoc:
+      ret = super
+      if ret.nil?
+        puts "CIAO #{default} of #{oid}"
+        ret = case default
+          # Numeric types are identified in a different way by cockroachdb
+          # postgresql adapter only looks for ::bigint, while cockroach
+          # has INT and DECIMAL.
+          when /\A\(?(-?\d+(\.\d*)?)\)?(:::(INT(4,8,16,32,64)?|DECIMAL))?\z/
+            $1
+          else
+            # Anything else is blank, some user type, or some function
+            # and we can't know the value of that, so return nil.
+            nil
+          end
+      end
+      ret
+    end
 end


lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -145,4 +145,26 @@ def table_alias_length
...
+  private
+
+  # Extracts the value from a PostgreSQL column default definition.
+  # Some of postgresql's default regexps are not going to work with cockroachdb,
+  # leaving the default values out of the schema.rb file.
+  def extract_value_from_default(oid, default) # :nodoc:
+    ret = super
+    if ret.nil?
+      ret = case default
+        # Numeric types are identified in a different way by cockroachdb
```
lego: Could you also put an example of a value that would fail before this into the comment? As a reader, it's not clear immediately what this regex is capturing.

(Only after comparing this to the postgres numeric regex, I know the difference here is INT vs. bignum)

lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -145,4 +145,26 @@ def table_alias_length
...
+        when /\A\(?(-?\d+(\.\d*)?)\)?(:::INT)?\z/
```
lego: A few remarks, you might be able to answer them as I have not looked into the details of how this works.
- Is this input string lowercased earlier on? If not, I might suggest adding the lowercase regex modifier here. I assume this because postgres uses `bigint` lower case.
- We also have sized types such as `INT{4,8,16,32,64}` and `DECIMAL`. Do we also want to extract the values from those? Considering this is just considered with extracting values, disregarding the types, I think this is appropriate.

flavio:
@lego: I addressed your concerns. Please take a look at the latest code I pushed.

lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -145,4 +145,29 @@ def table_alias_length
...
+
+  private
+
+    # Extracts the value from a PostgreSQL column default definition.
+    # Some of postgresql's default regexps are not going to work with cockroachdb,
+    # leaving the default values out of the schema.rb file.
+    def extract_value_from_default(oid, default) # :nodoc:
+      ret = super
+      if ret.nil?
+        puts "CIAO #{default} of #{oid}"
```
lego: You left a print here.

lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -145,4 +145,29 @@ def table_alias_length
...
+        ret = case default
+          # Numeric types are identified in a different way by cockroachdb
+          # postgresql adapter only looks for ::bigint, while cockroach
+          # has INT and DECIMAL.
+          when /\A\(?(-?\d+(\.\d*)?)\)?(:::(INT(4,8,16,32,64)?|DECIMAL))?\z/
```
lego: Thank you! 👏 

We also allow the use of `1::INT` (2 colons is a type cast, 3 is a type annotation). Could you change the last colon to `:?` condition for now.

BuonOmo:
I think this has now been handled differently. @rafiss I vote for closing.

PR #17, opened 2017-11-17
Disable some unsupported features

flavio:
Prevent AR from using features that are not yet supported by cockroachdb.


lego:
@flavio do know what disabling these will change in ActiveRecord? I'm just wondering in order to keep a record and know how it may impact others.

The referential integrity one is more straightforward. Disabling savepoints I would maybe expect some different behavior for transactions. I'll look into them if you aren't sure.

BuonOmo:
This is now outdated, thanks for contribuing!

Issue #24, opened 2018-04-05
Active Record 5.1 Failing to insert version number


bithavoc:
> ActiveRecord::StatementInvalid: PG::InternalError: ERROR:  value type int doesn't match type STRING of column "version"
: INSERT INTO "schema_migrations" (version) VALUES (20180402213849)
> PG::InternalError: ERROR:  value type int doesn't match type STRING of column "version"

```
show columns from schema_migrations;
+---------+--------+-------+---------+-------------+
|  Field  |  Type  | Null  | Default |   Indices   |
+---------+--------+-------+---------+-------------+
| version | STRING | false | NULL    | {"primary"} |
+---------+--------+-------+---------+-------------+
(1 row)

Time: 8.672796ms
```

knz:
@jordanlewis plz file the following separate issues in the lang project:

- [ ] insert int into string column (note: we may want to address this by pattern-matching the query instead of weird typing rule)
- [ ] implement `generate_subscripts`

bithavoc:
@knz this PR has the solution, can @jordanlewis review it? https://github.com/cockroachdb/activerecord-cockroachdb-adapter/pull/20/files#diff-a6efa568aab40b05159b3148c29a54f9R200

Cikey:
The PR #20 didn't solve the initial issue for me. I could not run any rails tests or anything that required schema loading.

Im using:
* cockroachdb `v2.0.1`
* cockroach adapter `0.2.3` (the readme still points to `0.2.2`)
* rails `5.2.0`

Looks like, this line is the error source: https://github.com/rails/rails/blob/73d1f5f310e4dc3acb1ea502143d679f99931840/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L1057

The schema version like `ActiveRecord::Schema.define(version: 2018_05_04_205847) do` is always cast to an int. When executing the insert statement, the int value is not inside single quotes. Postgres allows automatic int to varchar conversions that cockroach seem to prevent: [sqlfiddle](http://sqlfiddle.com/#!17/ea6a8/5/2)

If i remove the version block from the schema.rb file, the error disappears.

I resolved the issue by monkeypatching the cockroachdb_adapter and 'fix' the `assume_migrated_upto_version` method as proof of concept:

```
module CockroachDb
    module SchemaFix
      def assume_migrated_upto_version(version, migrations_paths)
        migrations_paths = Array(migrations_paths)
        version = version.to_i
[...]

PR #27, opened 2018-05-05
fix: schema load failing with int to string conversion error

Cikey:
I extracted the monkeypatch described [here](https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/24#issuecomment-386791078) into the existing CockrachDb::SchemaStatement module. 

This should solve #24 when executing `rails db:schema:load` or similar commands that involve schema.rb

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
@@ -52,6 +52,49 @@ module ActiveRecord
             query_value("SELECT setval(#{quote(quoted_sequence)}, #{max_pk ? max_pk : minvalue})", "SCHEMA")
           end
         end
+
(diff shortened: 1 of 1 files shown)


Cikey:
Quick heads up, i found that if more than one migration exists, it still fails in this branch of the code:

```
inserting = (versions - migrated).select { |v| v < version }
if supports_multi_insert?
  execute insert_versions_sql(inserting)
else
  inserting.each do |v|
    execute insert_versions_sql(v)
  end
end
``` 

I try to fix it tomorrow

Cikey:
This should fix the failing insert commands if more than one migration exists.

miry:
ping @jordanlewis on the problem to run `rails db:schema:load`

jordanlewis:
Thanks! I'll merge this as it's passing the test suite... curious that it wasn't passing before.

PR #31, opened 2018-09-25
Leverage semantic versioning.

cabello:
Given that the latest version is 0.2.3 suggest pinning version to allow bugfixes.


jordanlewis:
Thanks!!

PR #32, opened 2019-03-14
gemspec: Remove pg pinning

tonobo:
It isn't necessary to pin this gem anymore. The related commit
29175527605d58fd7f84b6cc2beb5096be0256ba from before already switches to
the new constant layout.


sandstrom:
@tonobo Seems to be solved already, maybe this can be closed?

https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/master/activerecord-cockroachdb-adapter.gemspec#L21

BuonOmo:
@rafiss it is indeed already taken care of. Could you close this PR?

PR #35, opened 2019-07-18
add Jets support

flippakitten:


jordanlewis:
What is Jets?

flippakitten:
@jordanlewis 
> What is Jets?

Ruby on Jets.
https://rubyonjets.com/
Basically it's Rails but runs in AWS Lambda's

I was contemplating a completely separate gem but as `Jets` uses active-record, it seems like a better option to include it in here.

BuonOmo:
@flippakitten are you still up to being champion for this PR?

If yes, I have a few questions.

Is it still worth it, as there are [rails support options](https://rubyonjets.com/docs/rails-support/) in jets?

I feel like this is not ok: we are basically loading the tasks every time. But in rails it is only done on task loading. There is a `Railtie` hook. I found a [`#rake_tasks`](https://rubydoc.info/gems/jets/Jets/Commands/RakeCommand#rake_tasks-class_method) method in jets, but not sure it works the same way.

Looking on [sourcegraph](https://sourcegraph.com/search?q=context:global+defined%3F%28Jets&patternType=standard&case=yes&sm=1&groupBy=repo), I think a railtie equivalent in jets is a turbine.

Anyway, there is a bit more investigation/work to do for this integration. Are you up to the task?

Otherwise I'd be in favor of closing as the output of sourcegraph has shown very few repos supporting jets, I don't think we should invest that much time in it!

<details>

also trains are better than jets <3

</details>

BuonOmo:
@flippakitten are you still up to being champion for this PR?

If yes, I have a few questions.

Is it still worth it, as there are [rails support options](https://rubyonjets.com/docs/rails-support/) in jets?

I feel like this is not ok: we are basically loading the tasks every time. But in rails it is only done on task loading. There is a `Railtie` hook. I found a [`#rake_tasks`](https://rubydoc.info/gems/jets/Jets/Commands/RakeCommand#rake_tasks-class_method) method in jets, but not sure it works the same way.

Looking on [sourcegraph](https://sourcegraph.com/search?q=context:global+defined%3F%28Jets&patternType=standard&case=yes&sm=1&groupBy=repo), I think a railtie equivalent in jets is a turbine.

Anyway, there is a bit more investigation/work to do for this integration. Are you up to the task?

Otherwise I'd be in favor of closing as the output of sourcegraph has shown very few repos supporting jets, I don't think we should invest that much time in it!

<details>

also trains are better than jets <3

</details>

PR #38, opened 2020-03-12
Run ActiveRecord tests from source instead of a fork

alimi:
Before this change, the ActiveRecord test suite was being run from a forked and modified version of Rails. This has worked well, but it's hard to maintain long term. As Rails changes, it'll get harder and harder to update the forked copy of Rails.

In this PR, I borrowed the [setup used by the ActiveRecord SQL Server Adapter](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/0a0ee519a423ac9a6a142834616a8e1a8252c7e7/RUNNING_UNIT_TESTS.md). Now we'll be able to run the ActiveRecord test suite against the Rails version specified by the adapter (currently Rails 5.2) by running `bundle exec rake test`. We'll also have the option to run against a local copy of Rails with `RAILS_SOURCE="path/to/local_copy" bundle exec rake test`. This will come in handy when debugging/prototyping. Finally, we can run specific ActiveRecord tests with `TEST_FILES_AR="path/to/test.rb,path/to/another_test.rb" bundle exec rake test`.

I updated the docs with this new setup and removed the Rails git submodule.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,11 +1,32 @@
+require 'openssl'
 source 'https://rubygems.org'
-
-# gem 'activerecord', git: 'https://github.com/lego/ruby-on-rails.git'
-
-git_source(:github) { |repo| "https://github.com/#{repo}.git" }
-
-# Specify your gem's dependencies in activerecord-cockroachdb-adapter.gemspec
 gemspec
 
-# We need a newish Rake since Active Job sets its test tasks' descriptions.
-gem "rake", ">= 11.1"
+if ENV['RAILS_SOURCE']
+  gemspec path: ENV['RAILS_SOURCE']
+else
+  def get_version_from_gemspec
+    require 'net/http'
+    require 'yaml'
+    spec = eval(File.read('activerecord-cockroachdb-adapter.gemspec'))
+    ver = spec.dependencies.detect{ |d|d.name == 'activerecord' }.requirement.requirements.first.last.version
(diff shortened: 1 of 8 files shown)


Gemfile
```diff
@@ -1,11 +1,39 @@
...
 
-git_source(:github) { |repo| "https://github.com/#{repo}.git" }
+if ENV['RAILS_SOURCE']
+  gemspec path: ENV['RAILS_SOURCE']
+else
+  def get_version_from_gemspec
+    require 'net/http'
+    require 'yaml'
+    spec = eval(File.read('activerecord-cockroachdb-adapter.gemspec'))
+    ver = spec.dependencies.detect{ |d|d.name == 'activerecord' }.requirement.requirements.first.last.version
```
apantel: Can this be done more cleanly?

apantel:
Looks good, just one point to clear up

Gemfile
alimi: Hmm...I'm not sure that it can but lemme give it a try.

alimi: I did a little refactor in https://github.com/cockroachdb/activerecord-cockroachdb-adapter/pull/38/commits/130bff27bde546b4686e9875cf27110ed6b98c80 to make it a little clearer what's going. Getting the version from the gemspec is still a little hairy because that's what the Rubygems API gives us, but the rest of the file should be easier to follow 😄.

PR #39, opened 2020-03-13
Update ReferentialIntegrity module

alimi:
`ReferentialIntegrity` defines `#disable_referential_integrity` to temporarily disable foreign key constraints. This is usually used to load test fixtures without having to worry about the order in which the data is inserted.

 Before this change the module would remove all the foreign key constraints, but it would only re-add them on the last table they were removed from.

https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/66a9c65dda8774a233149c9c1234faea8be1ce04/lib/active_record/connection_adapters/cockroachdb/referential_integrity.rb#L14-L18

`fkeys` was reset after every iteration of `tables.each`

https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/66a9c65dda8774a233149c9c1234faea8be1ce04/lib/active_record/connection_adapters/cockroachdb/referential_integrity.rb#L42-L45

and `fkeys` were the only keys that got re-added. Therefore, the database was left in a bad state missing most of it's foreign key constraints.

 `#disable_referential_integrity` also opened extra transactions that weren't needed. These transactions led to nested savepoints that CockroachDB doesn't currently support.

Now, all foreign key constraints that were removed will be re-added.

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb/referential_integrity.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/referential_integrity.rb
@@ -1,51 +1,36 @@
 # frozen_string_literal: true
-# NOTE(joey): This is cradled from connection_adapters/postgresql/referential_integrity.rb
-# It is commonly used for setting up fixtures during tests.
+
+# The PostgresSQL Adapter's ReferentialIntegrity module can disable and
+# re-enable foreign key constraints by disabling all table triggers. Since
+# triggers are not available in CockroachDB, we have to remove foreign keys and
+# re-add them via the ActiveRecord API.
+#
+# This module is commonly used to load test fixture data without having to worry
+# about the order in which that data is loaded.
 module ActiveRecord
   module ConnectionAdapters
     module CockroachDB
-      module ReferentialIntegrity # :nodoc:
-        def disable_referential_integrity # :nodoc:
-          original_exception = nil
-          fkeys = nil
+      module ReferentialIntegrity
+        def disable_referential_integrity
+          foreign_keys = tables.map { |table| foreign_keys(table) }.flatten
 
-          begin
-            transaction do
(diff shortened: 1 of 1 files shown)


lib/active_record/connection_adapters/cockroachdb/referential_integrity.rb
```diff
@@ -1,51 +1,36 @@
...
-
-Please go to https://github.com/cockroachdb/activerecord-cockroachdb-adapter
-and report this issue.
-
-    cause: #{original_exception.try(:message)}
-
-              WARNING
-            raise e
-          end
+          yield
```
apantel: Do you not still want a rescue on the yield? It seems like you still want to re-add the FKs if the block errors.

apantel:
One comment, otherwise good

lib/active_record/connection_adapters/cockroachdb/referential_integrity.rb
alimi: Good question! The `rescue` wouldn't re-add the foreign keys if the `yield` errored. It only captured `ActiveRecord::InvalidForeignKey` errors, and even for those errors the `rescue` re-raised the error causing execution to stop.

The `rescue` was [copied from the PostgreSQL adapter](https://github.com/rails/rails/blob/8bec77cc0f1fd47677a331a64f68c5918efd2ca9/activerecord/lib/active_record/connection_adapters/postgresql/referential_integrity.rb#L18-L31), and it's there to catch permissions problems with disabling foreign keys. Since we're removing foreign keys via the AR API, I don't think we'll run into the same problem. Does that make sense?

apantel: Makes sense, thanks!

PR #41, opened 2020-03-19
Retry transactions that fail from serialization errors 

alimi:
ActiveRecord makes heavy use of savepoints to support nested transactions. That was disabled in the adapter because CockroachDB didn't support nested savepoints, but as of v20.1 it does 🎉

Instead of disabling nested transactions, the adapter can now retry transactions that error due to serialization errors. Since [ActiveRecord will rollback these erroring transactions](https://github.com/rails/rails/blob/ac30e389ecfa0e26e3d44c1eda8488ddf63b3ecc/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb#L237-L245), the adapter only needs to manage the number of retries.

The number of retries is configurable. Users will be able set it in `config/database.yml` like

```yml
  development:
    <<: *default
    database: app_development
    max_transaction_retries: 5
```

The current default retries of 3 was an arbitrary choice so lemme know if we should change it to something else.

I also moved `transaction_isolation_levels` to a new module, `CockroachDB::DatabaseStatements`. This will keep the adapter more in line with the other ActiveRecord adapters, and it'll be easier find what's been changed.

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb/transaction_manager.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/transaction_manager.rb
@@ -15,7 +15,7 @@ module ActiveRecord
           raise if attempts >= @connection.max_transaction_retries
 
           attempts += 1
-          sleep_seconds = (((2 ** attempts) * 100) + rand( 100 - 1 ) + 1) / 1000.0
+          sleep_seconds = (2 ** attempts + rand) / 10
           sleep(sleep_seconds)
           within_new_transaction(options.merge(attempts: attempts)) { yield }
         end
--- a/lib/active_record/connection_adapters/cockroachdb/transaction_manager.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/transaction_manager.rb
@@ -1,25 +1,23 @@
 # frozen_string_literal: true
 
-require 'active_record/connection_adapters/abstract/transaction'
-
 module ActiveRecord
   module ConnectionAdapters
-
-    # NOTE(joey): This is a very sad monkey patch. Unfortunately, it is
-    # required in order to prevent doing more than 2 nested transactions
-    # while still allowing a single nested transaction. This is because
-    # CockroachDB only supports a single savepoint at the beginning of a
-    # transaction. Allowing this works for the common case of testing.
     module CockroachDB
       module TransactionManagerMonkeyPatch
-        def begin_transaction(options={})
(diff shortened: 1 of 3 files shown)


lib/active_record/connection_adapters/cockroachdb/transaction_manager.rb
```diff
@@ -1,25 +1,23 @@
...
+        # transactions will be retried until they pass or the max retry limit is
+        # exceeded.
+        def within_new_transaction(options = {})
+          attempts = options.fetch(:attempts, 0)
+          super
+        rescue ActiveRecord::SerializationFailure => error
+          raise if attempts >= @connection.max_transaction_retries
+
+          attempts += 1
+          sleep_seconds = (((2 ** attempts) * 100) + rand( 100 - 1 ) + 1) / 1000.0
```
apantel: Why is this random?

alimi: It's based on the example from the docs: https://www.cockroachlabs.com/docs/v19.2/transactions.html#client-side-intervention-example.

alimi:
I cleaned up the merge conflicts.

lib/active_record/connection_adapters/cockroachdb/transaction_manager.rb
apantel: I guess the idea is to desync if there are two calls that start at the same time. Clean up the math a bit and it's good.

alimi: What should we change the math to? It's currently the same math that's used in the example.

```python
# This is a retry error, so we roll back the current transaction
# and sleep for a bit before retrying. The sleep time increases
# for each failed transaction.  Adapted from
# https://colintemple.com/2017/03/java-exponential-backoff/
conn.exec('ROLLBACK');
sleep_ms = int(((2**n) * 100) + rand( 100 - 1 ) + 1)
```

apantel: I'd prefer `(2 ** attempts + rand()) / 10`. It's also not clear to me why the +1 is there.

alimi: Yep, that's a lot cleaner and gets us the same results. Updated in 64d450f5007f59a55f1431be0532ca4d294c75c9.

PR #43, opened 2020-03-24
Remove overridden indexes method

alimi:
`#indexes` was overridden in the adapter because the PostgreSQL Adapter definition didn't work with CockroachDB. Now that's no longer true, so we can remove the overridden method 🎉

This also includes some changes to the test setup. I brought in minitest-excludes so we can easily skip ActiveRecord tests that aren't compatible with CockroachDB (this approach is borrowed from the [ActiveRecord JDBC (JRuby) Adapter](https://github.com/jruby/activerecord-jdbc-adapter/tree/90249328615fd0cb5c2d9b61bc659b603c128cc4/test/rails/excludes)). When appropriate, we can re-write the excluded tests so they work with CockroachDB.

For example, [`test/excludes/ActiveRecord/AdapterTest.rb`](https://github.com/alimi/activerecord-cockroachdb-adapter/blob/a4bca1bac4b009b731bdcb4a3708c3c06be679c8/test/excludes/ActiveRecord/AdapterTest.rb) skips a couple of index tests that were being run inside of transactions. [`test/cases/adapter_test.rb`](https://github.com/alimi/activerecord-cockroachdb-adapter/blob/a4bca1bac4b009b731bdcb4a3708c3c06be679c8/test/cases/adapter_test.rb) brings the tests back but runs them without transactions so they work against CockroachDB.

There are also more test flags that can be used such as `ONLY_TEST_AR` and `ONLY_TEST_COCKROACHDB` to only run the ActiveRecord test suite or CockroachDB Adapter test suite respectively. The debug flag, `DEBUG_COCKROACHDB_ADAPTER`, is also turned on by default in the test environment.

**NOTE: this PR is based off of master, but the changes will only work once #38 and #40 are merged**


alimi:
Now that #38 and #40 have been merged, this should be good to go.

PR #46, opened 2020-03-31
Remove overridden reset_pk_sequence!

alimi:
This removes the overridden `reset_pk_sequence!` method in favor of  the PostgreSQL adapter's definition.

This also defines `postgresql_version` and sets it to the 100000 (version 10.00.00). ActiveRecord checks for this as the minimum version before deciding which queries to run. So far, it appears PostgreSQL 10 queries work against CockroachDB.


test/cases/fixtures_test.rb
```diff
@@ -0,0 +1,152 @@
...
+      Account.connection.create_table :accounts, force: true do |t|
+        t.references :firm, index: false
+        t.string  :firm_name
+        t.integer :credit_limit
+      end
+
+      Company.connection.drop_table :companies, if_exists: true
+      Company.connection.exec_query("DROP SEQUENCE IF EXISTS companies_nonstd_seq")
+      Company.connection.create_table :companies, force: true do |t|
+        t.string  :type
```
alimi: In these tests, we're dropping tables, recreating them with primary key sequences, and bringing back the original tables. This _could_ lead to problems because we're bringing back the tables based on definitions that _could_ change in future versions of Rails. I don't think that's likely, but if it happens we can create our own tables with primary key sequences and leave the Rails' ones alone.

Even though this is a little janky, it keeps us pretty closely aligned to the original AR tests.

alimi:
I wonder how often `reset_pk_sequence!` will be used in real world apps. AR will default to creating tables with serial primary keys, and serial primary keys are backed by `unique_rowid()` in CockroachDB. `reset_pk_sequence!` will most likely be a no-op for most tables because there won't be any sequences to reset.

Maybe people will have tables with primary key sequences if they're migrating from another db like PostgreSQL?

PR #56, opened 2020-04-06
Skip more schema tests

alimi:
Most of these schema tests are skipped because custom schemas are not supported in CockroachDB (https://github.com/cockroachdb/cockroach/issues/26443).

We're skipping a few more tests for different reasons
  1. NULLS FIRST and NULLS LAST aren't implemented in CockroachDB (https://www.cockroachlabs.com/docs/v19.2/null-handling.html#nulls-and-sorting)
  1. Operator classes on indexes are not supported in CockroachDB. (https://www.postgresql.org/docs/9.5/indexes-opclass.html)

I wasn't able to find any documentation or issues on operator classes. Is this documented anywhere or is it a new problem?


apantel:
It looks like nobody has asked for operator class support in cockroachdb. Can you file an issue and point to it, or tell me what it should say?

alimi:
I opened https://github.com/cockroachdb/cockroach/issues/47420. I wasn't sure how to label it and stuff, so let me know if it needs any changes.

apantel:
Looks good, thanks!

PR #60, opened 2020-04-08
Replace some LeftOuterJoinAssociationTest tests

alimi:
One test failed because it assumed the query result order would also be the same. Another failed trying to do string concatenation with non-strings. Both have been fixed to work against CockroachDB.


test/excludes/LeftOuterJoinAssociationTest.rb
```diff
@@ -0,0 +1,2 @@
+exclude :test_construct_finder_sql_applies_aliases_tables_on_association_conditions, "The test fails because the query result order is not guaranteed."
```
apantel, line 1: This won't stop the version you wrote about from running, right?

alimi: It'll prevent the ActiveRecord version of this test from running, but it won't stop `test/cases/associations/left_outer_join_association_test.rb ` from running. Does that answer your question?

apantel: Yes, thanks!

PR #61, opened 2020-04-08
Override ActiveRecord::Type#current_adapter_name

alimi:
Instead of letting the original definition infer the `current_adapter_name` as `:cockroachdb`, force it to return `:postgresql` so we can continue to use [the `ActiveRecord::Type`s defined in `PostgreSQLAdapter`](https://github.com/rails/rails/blob/ac30e389ecfa0e26e3d44c1eda8488ddf63b3ecc/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L842-L860).


alimi:
Fixed the merge conflict so this should be good to merge 🚀

PR #62, opened 2020-04-10
ActiveRecord base test fixes

alimi:
This includes a few changes to fix [ActiveRecord's base test](https://github.com/rails/rails/blob/ac30e389ecfa0e26e3d44c1eda8488ddf63b3ecc/activerecord/test/cases/base_test.rb).

The biggest change is the introduction of a CockroachDB specific test schema. It's a copy of the PostgreSQL test schema defined by ActiveRecord with a few changes so it works for CockroachDB.

See the commits for more details on the changes.

Note: there's an outstanding failure in the base test. [`test_marshal_between_processes`](https://github.com/rails/rails/blob/ac30e389ecfa0e26e3d44c1eda8488ddf63b3ecc/activerecord/test/cases/base_test.rb#L1187-L1216) causes a Ruby segfault when it tries to open a new connection in a forked process. I'm not sure what's going there so I'm tabling it for now.

### DIFF
--- a/test/cases/base_test.rb
+++ b/test/cases/base_test.rb
@@ -1,0 +1,31 @@
+require "cases/helper_cockroachdb"
+
+module CockroachDB
+  class BasicsTest < ActiveRecord::TestCase
+    # This replaces the same test that's been excluded from BasicsTest. It's
+    # exactly the same, except badchar has an entry for CockroachDBAdapter.
+    def test_column_names_are_escaped
+      conn      = ActiveRecord::Base.connection
+      classname = conn.class.name[/[^:]*$/]
+      badchar   = {
+        "SQLite3Adapter"    => '"',
+        "Mysql2Adapter"     => "`",
+        "PostgreSQLAdapter" => '"',
+        "OracleAdapter"     => '"',
+        "FbAdapter"         => '"',
+        "CockroachDBAdapter" => '"'
+      }.fetch(classname) {
+        raise "need a bad char for #{classname}"
+      }
+
+      quoted = conn.quote_column_name "foo#{badchar}bar"
+      if current_adapter?(:OracleAdapter)
+        # Oracle does not allow double quotes in table and column names at all
+        # therefore quoting removes them
+        assert_equal("#{badchar}foobar#{badchar}", quoted)
+      else
+        assert_equal("#{badchar}foo#{badchar * 2}bar#{badchar}", quoted)
+      end
+    end
+  end
+end
--- a/test/schema/cockroachdb_specific_schema.rb
+++ b/test/schema/cockroachdb_specific_schema.rb
@@ -1,0 +1,122 @@
+# This is a copy of postgresql_specific_schema.rb from ActiveRecord's test
+# suite. Statements that don't work against CockroachDB have been commented out
+# with an explanation of why they don't work.
+
+ActiveRecord::Schema.define do
+
+  enable_extension!("uuid-ossp", ActiveRecord::Base.connection)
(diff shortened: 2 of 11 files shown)


alimi:
The Ruby segfault problem might be specific to the pg gem on macOS. Per https://github.com/ged/ruby-pg/issues/311, I could get to the test to run and pass with `gssencmode: disable` in my `database.yml`. The test still segfaulted when I ran it with other tests though. We'll have to see if this problem exists in other environments like CI.

h/t @marlabrizel for having seen this problem before 🌟

alimi:
Forgot to commit f110047 on Friday.

test/schema/cockroachdb_specific_schema.rb
```diff
@@ -0,0 +1,122 @@
...
+  # end
+
+  execute <<_SQL
+  CREATE TABLE postgresql_timestamp_with_zones (
+    id SERIAL PRIMARY KEY,
+    time TIMESTAMP WITH TIME ZONE
+  );
+_SQL
+
+# Since stored procedures are not supported in CockroachDB, it won't be possible
```
apantel, line 73: Should this be removed?

test/cases/base_test.rb
```diff
@@ -0,0 +1,31 @@
...
+
+module CockroachDB
+  class BasicsTest < ActiveRecord::TestCase
+    # This replaces the same test that's been excluded from BasicsTest. It's
+    # exactly the same, except badchar has an entry for CockroachDBAdapter.
+    def test_column_names_are_escaped
+      conn      = ActiveRecord::Base.connection
+      classname = conn.class.name[/[^:]*$/]
+      badchar   = {
+        "SQLite3Adapter"    => '"',
```
apantel, line 11: If this lookup table exist in the base, should the entry for CRDB be there?

test/schema/cockroachdb_specific_schema.rb
alimi: Since this file is a copy of AR's PostgreSQL specific schema, I left stuff like this in place but commented out thinking it might make it easier to see the differences between the two files. We can take it out if it isn't helpful. What do you think?

test/cases/base_test.rb
alimi: Good question! The lookup table is [in the ActiveRecord test suite](https://github.com/rails/rails/blob/a0fb2b2227c7e9eacceac65b40c8a59dbfb497ec/activerecord/test/cases/base_test.rb#L100-L108), but I'm not sure they'll accept a PR to add CockroachDB since it's not an officially supported adapter.

test/schema/cockroachdb_specific_schema.rb
apantel: Yeah, this is ok to keep.

apantel:
Looks good, just making sure the branch is in the state you want.

alimi:
It's ready to merge. :shipit:

alimi:
I added a couple more changes to fix things I came across in `test/cases/defaults_test.rb`.

https://github.com/cockroachdb/activerecord-cockroachdb-adapter/pull/62/commits/7ad6f5357643923bfafddbd68c3dd0170d6b6e4a updates `extract_value_from_default` so it handles escaped strings. See https://github.com/cockroachdb/cockroach/issues/47497.

https://github.com/cockroachdb/activerecord-cockroachdb-adapter/pull/62/commits/98bf1a8abd7be6d7147f0568c40e81e6d0c4cc0e replaces an AR test to work with CockroachDB's `current_date()` and `current_timestamp()` functions.

PR #65, opened 2020-04-14
Update CI scripts

rafiss:
With these changes, the CI job on TeamCity should start workign again.

We still need to fix the script so it runs all the tests without
crashing. Currently, it only runs two of the test cases.


alimi:
Is the script crashing because tests are crashing, or is it crashing for other reasons?

rafiss:
It was hanging when running all the tests, but it seemed to be working when I used the latest release candidate version of 20.1.

PR #67, opened 2020-04-14
Move gem deps for AR test suite to dev group

alimi:
Some gems are needed to run the ActiveRecord test suite, but they're not required to use the adapter. We've had these gems listed in a few places, and now we'll consolidate them under the Gemfile's development group.

This PR also adds sqlite3 as another dep for the ActiveRecord test suite.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,8 +2,6 @@ require 'openssl'
 source 'https://rubygems.org'
 gemspec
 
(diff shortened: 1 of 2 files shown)


alimi:
This change fixes at least a couple of tests:  `test/cases/connection_adapters/connection_specification_test.rb` and `test/cases/connection_adapters/merge_and_resolve_default_url_config_test.rb`.

PR #82, opened 2020-04-17
Run test cases separately in CI

rafiss:
It seems that the CI build is hanging. In order to debug this, we now
run each test separately.

cc @alimi @marlabrizel


build/teamcity-test.sh
```diff
@@ -41,19 +41,24 @@ bundle install
...
-    HAS_FAILED=1
-else
-    echo "Tests passed"
-    HAS_FAILED=0
-fi
+source ./build/test_cases.sh
+
+HAS_FAILED=0
+for test_file in ${TEST_CASES[@]}; do
+    if ! (TEST_FILES_AR="$test_file" bundle exec rake test); then
```
rafiss, line 48: Hmm actually does this work? Based on the instructions in #76 it seems like for the tests that we had to add/modify, it should be specified with `TEST_FILES` instead?

alimi: `TEST_FILES_AR` will run the ActiveRecord tests, but you're right--it won't include the ones we've changed. Maybe we can add another loop that goes through `test/cases/**/*test.rb` and sets `TEST_FILES`. Does that make sense?

rafiss:
Now we know that `test_raises_Deadlocked_when_a_deadlock_is_encountered` in test/cases/adapters/postgresql/transaction_test.rb was the cause of hanging builds, so we don't need this.

PR #94, opened 2020-04-17
Skip AR Transaction LockWaitTimeout test

alimi:
The test tries to set `lock_timeout`, but `lock_timeout` is not supported by CockroachDB.


test/excludes/ActiveRecord/PostgresqlTransactionTest.rb
```diff
@@ -0,0 +1 @@
+exclude :test_raises_LockWaitTimeout_when_lock_wait_timeout_exceeded, "The test tries to set lock_timeout, but lock_timeout is not supported by CockroachDB."
```
rafiss, line 1: thanks! i think we should also exclude `test "raises QueryCanceled when canceling statement due to user request"`

because cockroachdb doesn't support `pg_cancel_backend()`.

alimi: 👍🏾 - skipped it in https://github.com/cockroachdb/activerecord-cockroachdb-adapter/pull/94/commits/324620de68ec6cf9b1a5a0243b4fbc1eca4f134f

rafiss:
some tests are still hanging, but this is good to merge

PR #113, opened 2020-04-20
Replace OrTest#test_or_when_grouping

### LINKED ISSUE #87: [ActiveRecord 5.2 Test Failure] test/cases/relation/or_test.rb
```
Failure:
ActiveRecord::OrTest#test_or_when_grouping [/Users/alimi/repos/rails/activerecord/test/cases/relation/or_test.rb:95]:
--- expected
+++ actual
@@ -1 +1 @@
-[["hello", 6], ["Such a lovely day", 1]]
+[["Such a lovely day", 1], ["hello", 6]]


bin/rails test Users/alimi/repos/rails/activerecord/test/cases/relation/or_test.rb:92
```

See the [Contributing Guide](https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/master/CONTRIBUTING.md#setup-and-running-tests) for instructions on running tests.

If the test isn't valid against CockroachDB

1. Add a ruby file to `test/excludes` that matches the name of the test class if one doesn't already exist. For example to exclude a test from `ActiveRecord::AdapterTest`, create `test/excludes/ActiveRecord/AdapterTest.rb`.
1. Add an exclude statement to the file with the name of the test to exclude and a description. For example to exclude `test_indexes` from `ActiveRecord::AdapterTest`:
  https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/c3cb637bd855b36bbf9e76b7947162054b95022e/test/excludes/ActiveRecord/AdapterTest.rb#L1
1. Finally, if the test can run against CockroachDB with a few changes, add it to `test/cases`. Everything from the ActiveRecord test suite will be available, so a lot of the excluded test can be copied over. Namespace the test under the `CockroachDB` module to avoid name collisions. See [`test/cases/adapter_test.rb`](https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/6a0a02e9a54eaf6c1145501dea78cb54f8bb6804/test/cases/adapter_test.rb) for an example.

See https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/48.

marlabrizel:
Replaces `OrTest#test_or_when_grouping` due to non-deterministic test output. 

This has been [fixed upstream in Rails](https://github.com/rails/rails/pull/38978); the fix is replicated here to cover our needs.

Closes #87

### DIFF
--- a/test/cases/relation/or_test.rb
+++ b/test/cases/relation/or_test.rb
@@ -1,0 +1,26 @@
+# This test has been copied to fix a bug in the test setup. It can be removed
+# once the bugfix has been released in Rails - see rails/rails#38978.
+# See test/excludes/OrTest.rb
+
+require "cases/helper_cockroachdb"
+
+# Load dependencies from ActiveRecord test suite
+require "cases/helper"
+require "models/post"
+require "models/author"
+require "models/categorization"
+
+module CockroachDB
+  module ActiveRecord
+    class OrTest < ActiveRecord::TestCase
+      fixtures :posts
+      fixtures :authors, :author_addresses
+
+      def test_or_when_grouping
+        groups = Post.where("id < 10").group("body")
+        expected = groups.having("COUNT(*) > 1 OR body like 'Such%'").count
+        assert_equal expected, groups.having("COUNT(*) > 1").or(groups.having("body like 'Such%'")).count
+      end
+    end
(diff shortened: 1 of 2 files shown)


apantel:
This looks fine once the CI works.

PR #129, opened 2020-05-19
Run test cases separately in CI

rafiss:
Try out https://github.com/cockroachdb/cockroach/pull/48608 to see if it makes this test run faster. (Previously took 2+ days)


rafiss:
This run took 2 hours, so looks like a big performance improvement. (The test failure is expected given the current state of this repo.) Closing this out since we got the data we hoped for.

cc @spaskob

PR #134, opened 2020-09-02
Enable enum and partial index tests

rafiss:
Add a version check to enable partial indexes if >=20.2, and test with v20.2


test/cases/migration/index_test.rb
```diff
@@ -0,0 +1,199 @@
+# frozen_string_literal: true
+
+require "cases/helper"
+
+module ActiveRecord
+  module CockroachDB
+    class Migration
+      class IndexTest < ActiveRecord::TestCase
+        # These tests are identical to the ones found in Rails, save for the fact
```
angelapwen, line 9: Just for my own knowledge: how are you turning off the tests related to transactions here? Did you just remove those lines?

lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -83,8 +83,7 @@ def supports_materialized_views?
       end
 
       def supports_partial_index?
-        # See cockroachdb/cockroach#9683
-        false
+        cockroachdb_version >= 202
```
angelapwen, line 86: I think you can just do `@crdb_version >= 202` here and forego the `cockroachdb_version` method entirely.

angelapwen (approved):
👍 thanks for walking us through this PR today! Just a couple questions/nits that I'm curious about, but looks good to 🚢 otherwise

test/cases/migration/index_test.rb
rafiss: the self.use_transactional_tests setting turns off transactions, and also i only copied in the tests that are expected to pass

lib/active_record/connection_adapters/cockroachdb_adapter.rb
rafiss: that seems right. i just did it this way since i saw that's what the postgres_adapter does (https://github.com/rails/rails/blob/2a3f8a29e5d7d17a0448fa47fd756c8d11b175b1/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L400)

but don't think there's a particular reason to do so

test/cases/migration/index_test.rb
angelapwen: Oh I see. Didn't realize you could just use that built-in line to turn off transactions. Found docs: https://edgeguides.rubyonrails.org/testing.html#testing-parallel-transactions

rafiss:
thanks for the review!

PR #137, opened 2020-11-05
[WIP] Rails 6 update

albarnaz:
This is a feature branch that updates this adapter to use Rails 6.


rafiss:
Looks good! As expected, there are failing tests. Should we keep this on a separate branch so we can keep merging changes before fixing the tests all at the same time?

albarnaz:
@rafiss yes, the idea is to keep this as a feature branch, we will fix the tests and merge on this branch

activerecord-cockroachdb-adapter.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |spec|
   spec.description   = "Allows the use of CockroachDB as a backend for ActiveRecord and Rails apps."
   spec.homepage      = "https://github.com/cockroachdb/activerecord-cockroachdb-adapter"
 
-  spec.add_dependency "activerecord", "~> 5.2"
+  spec.add_dependency "activerecord", "~> 6.0"
```
carloslopes, line 16: Like we spoke through Slack, we will work based on the latest version of the active record which is 6.1

Cikey:
I checked our your work for one of my clients projects and found a bug with the retry logic: The monkey patch is not matching the correct method signature, it was changed in [this commit](https://github.com/rails/rails/commit/05c718a109184440ecc3e254e13f1145c30b2a6c#diff-dd0ea9dfea1fd486f2b1b5eac8c798f839467173a1c2315d3ad04f4098f7dda3R292). We are using activerecord 6.0.3.4 . 

I build a fixed version [here](https://github.com/Cikey/activerecord-cockroachdb-adapter/commit/4d0f10e471131b558232ce4ed9a2166ee7a5851d):

```ruby
module ActiveRecord
  module ConnectionAdapters
    module CockroachDB
      module TransactionManagerMonkeyPatch
        # Capture ActiveRecord::SerializationFailure errors caused by
        # transactions that fail due to serialization errors. Failed
        # transactions will be retried until they pass or the max retry limit is
        # exceeded.
        def within_new_transaction(isolation: nil, joinable: true, attempts: 0)
          super
        rescue ActiveRecord::StatementInvalid => error
          raise unless retryable? error
          raise if attempts >= @connection.max_transaction_retries

          attempts += 1
          sleep_seconds = (2 ** attempts + rand) / 10
          sleep(sleep_seconds)
          within_new_transaction(isolation: isolation, joinable: joinable, attempts: attempts) { yield }
        end

        def retryable?(error)
          return true if error.is_a? ActiveRecord::SerializationFailure
          return retryable? error.cause if error.cause
          false
        end
      end
    end

    class TransactionManager
      prepend CockroachDB::TransactionManagerMonkeyPatch
    end
  end
end
```

The ActiveRecord::SerializationFailure seems to be wrapped in a ActiveRecord::StatementInvalid.

rafiss:
We'll merge this to make a beta release, but will address the retry logic before doing the final release

PR #138, opened 2020-11-10
[Fix] Relation test override

albarnaz:
This PR override 2 tests from Relation tests: `test_relation_with_annotation_includes_comment_in_to_sql` and `test_relation_with_annotation_filters_sql_comment_delimiters`

The activerecord tests are expecting that `Post.where(id: 1)` returns `... WHERE id = 1`, but with a quote feature from cockroachdb it returns `... WHERE id = '1'`.

This behavior is already in the old version in production, so it will not raise an error.

### DIFF
--- a/test/excludes/RelationTest.rb
+++ b/test/excludes/RelationTest.rb
@@ -1,1 +1,3 @@
 exclude :test_finding_with_sanitized_order, "Skipping until we can triage further. See https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/48"
+exclude :test_relation_with_annotation_filters_sql_comment_delimiters, "Skipping test because the new feature from rails 6 annotate, this dont raise any error"
+exclude :test_relation_with_annotation_includes_comment_in_to_sql, "Skipping test because the new feature from rails 6 annotate, this dont raise any error"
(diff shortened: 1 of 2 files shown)


rafiss (approved):
This looks good. Should we merge this into the `feature/rails-6-update` branch now? It might be better to merge it now that we don't have to wait for all the tests to be fixed.

test/excludes/RelationTest.rb
```diff
@@ -1 +1,3 @@
 exclude :test_finding_with_sanitized_order, "Skipping until we can triage further. See https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/48"
+exclude :test_relation_with_annotation_filters_sql_comment_delimiters, "Skipping test because the new feature from rails 6 annotate, this dont raise any error"
```
rafiss, line 2: minor request: could the reason be rephrased to

"This test is overridden for CockroachDB because this adapter adds quotes to numeric values."

test/excludes/RelationTest.rb
```diff
@@ -1 +1,3 @@
...
+exclude :test_relation_with_annotation_includes_comment_in_to_sql, "Skipping test because the new feature from rails 6 annotate, this dont raise any error"
```
rafiss, line 3: same here:

"This test is overridden for CockroachDB because this adapter adds quotes to numeric values."

albarnaz:
@rafiss yes, I will resolve your comments and then we merge it in `feature/rails-6-update`

PR #141, opened 2020-11-18
[Fix] Adapter test foreign key verification

albarnaz:
This PR overrides the foreign key test to prevent an intermittent error.
During the tests, the table fk_test_has_pk has the constrain dropped and not created back.


test/excludes/ActiveRecord/AdapterForeignKeyTest.rb
```diff
@@ -1,2 +1,4 @@
 exclude :test_foreign_key_violations_are_translated_to_specific_exception_with_validate_false, "Skipping until we can triage further. See https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/48"
 exclude :test_foreign_key_violations_are_translated_to_specific_exception, "Skipping until we can triage further. See https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/48"
```
rafiss, line 2: can you override these two tests to see if they pass?

rafiss (changes requested):

test/excludes/ActiveRecord/AdapterForeignKeyTest.rb
albarnaz: One of these tests are completely removed from the test suit, the other one I override

rafiss (approved):

PR #148, opened 2020-11-27
[Fix]Fixture test sequence fix

albarnaz:
This PR moves some tests from ActiveRecord to Cockroach tests suite.
The tests are exactly the same.
Fix the connection to use Cockroach base.
Fix the primary sequence to use fixtures and don't use CockroachDB's `unique_rowid()`


test/excludes/FixturesTest.rb
```diff
@@ -1,2 +1,11 @@
 exclude :test_bulk_insert_multiple_table_with_a_multi_statement_query, "Skipping until we can triage further. See https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/48"
 exclude :test_bulk_insert_with_a_multi_statement_query_in_a_nested_transaction, "Skipping until we can triage further. See https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/48"
+exclude :test_clean_fixtures, "Skipping"
+exclude :test_auto_value_on_primary_key, "Skipping"
```
rafiss, line 4: can you make the comment say "Skipping the PostgreSQL test, but reimplemented for CockroachDB in test/cases/fixtures_test.rb"

rafiss (changes requested):
nice, glad these got fixed!

test/excludes/FixturesTest.rb
albarnaz: Done

PR #149, opened 2020-12-01
[Fix] Fixture test get sequence

albarnaz:
This PR overrides the reset_pk_sequence from schema_statement.

This method needs to be overridden to run the right query to get the minimum value for a sequence.

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
@@ -62,6 +62,7 @@ module ActiveRecord
           sql
         end
 
+        # This overrides the method from PostegreSQL adapter
         # Resets the sequence of a table's primary key to the maximum value.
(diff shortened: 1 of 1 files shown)


lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
```diff
@@ -57,10 +57,34 @@ def type_to_sql(type, limit: nil, precision: nil, scale: nil, array: nil, **) #
             end
           # The call to super might have appeneded [] already.
           if array && type != :primary_key && !sql.end_with?("[]")
-            sql = "#{sql}[]" 
+            sql = "#{sql}[]"
           end
           sql
         end
+
+        # Resets the sequence of a table's primary key to the maximum value.
```
rafiss, line 65: Can you update the comment to say that this overrides the method from the Postgres adapter?

rafiss (changes requested):

rafiss (approved):

PR #160, opened 2020-12-17
Update retry logic for ActiveRecord 6

rafiss:
Also, add a check so it only retries on SerializationFailure.

Thanks to @Cikey for the patch: https://github.com/Cikey/activerecord-cockroachdb-adapter/commit/4d0f10e471131b558232ce4ed9a2166ee7a5851d


solongordon (approved):

Cikey:
Its great to see rails 6 support, thanks!

PR #161, opened 2020-12-18
Fix database_exists test

rafiss:


lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -24,12 +24,9 @@ def cockroachdb_connection(config)
       valid_conn_param_keys = PG::Connection.conndefaults_hash.keys + [:sslmode, :application_name]
       conn_params.slice!(*valid_conn_param_keys)
 
-      # The postgres drivers don't allow the creation of an unconnected
```
angelapwen, line 27: Just curious, are these comments no longer required because we should always have a connected object now?

angelapwen (approved):
Just a quick question but good to go! 🚀

lib/active_record/connection_adapters/cockroachdb_adapter.rb
rafiss: yeah exactly!  i noticed that the `conn = PG.connect(conn_params)` line was just added a few weeks ago, and before that a nil value was being passed in as the first argument. the commend was referring to that nil argument but should have been removed earlier.

rafiss:
thanks for reviewing!

lib/active_record/connection_adapters/cockroachdb_adapter.rb
angelapwen: Got it, I was trying to understand how the comment's removal related to the rest of the changes in this PR. That makes sense!

PR #165, opened 2021-01-06
Override test_finding_with_subquery_with_eager_loading_in_from

rafiss:
The test relies on deterministic order of SELECT results, but did not
include an ORDER BY


otan (approved):

solongordon:
Is this a change we could contribute back to Active Record?

rafiss:
yeah we could -- i'll try that out.

PR #174, opened 2021-01-20
Add back quoting to schema migration version

### LINKED ISSUE #168: shema_migration use int instead of varchar
I'm using rails 6 and the new cockorach adapter (6.0.0beta2)

When I run `rails db:schema:load` (which is used in many rails tasks like db:migrate or db:setup) I have the following error: 
```
PG::DatatypeMismatch: ERROR:  value type int doesn't match type varchar of column "version"
```

This is due to rails which convert the migration version value in an `Int` and try to insert in a `varchar` column in Postgres.
Postgres seem to be tolerant and finally convert it again in varchar.
But this is not the case with cockroach. 

this commit 
https://github.com/cockroachdb/activerecord-cockroachdb-adapter/commit/5247400892e5ec9df4c837e3399bc1b449121f27#diff-c8c37bcec27b647173b8313d65f1724a297c9792dff5d8aae580443f1d78636d
patched it.

But it was reverted here:
https://github.com/cockroachdb/activerecord-cockroachdb-adapter/commit/2deabbf8847ebb429c45d4490d163b0709bdf248#diff-c8c37bcec27b647173b8313d65f1724a297c9792dff5d8aae580443f1d78636d

I guess we should re-add this patch

rafiss:
This was reverted in 08c5ef7b65edaa9be090345ab874a46890f7a13e
but we still need it

fixes #168

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
@@ -68,6 +68,51 @@ module ActiveRecord
           )
         end
 
+        # copied from ConnectionAdapters::SchemaStatements
+        #
+        # modified insert into statement to always wrap the version value into single quotes for cockroachdb.
+        def assume_migrated_upto_version(version, migrations_paths = nil)
+          unless migrations_paths.nil?
+            ActiveSupport::Deprecation.warn(<<~MSG.squish)
+            Passing migrations_paths to #assume_migrated_upto_version is deprecated and will be removed in Rails 6.1.
+            MSG
+          end
+
+          version = version.to_i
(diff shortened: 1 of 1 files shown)


solongordon (approved):
Could the commit message say more about why this is still needed?

rafiss:
@solongordon actually, i think i have a better understanding of the original bug. it could be related to an unexpected quirk of the rails configuration causing the `postgresql` adapter to be used instead of the `cockroachdb` one. See https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/168#issuecomment-764857777

let's see if that solves the user's original bug report.

rafiss:
not needed. see #168

PR #179, opened 2021-02-10
gemspec: remove open ended dependency

otan:
Got
```
WARNING:  open-ended dependency on pg (>= 1.2) is not recommended
  if pg is semantically versioned, use:
    add_runtime_dependency 'pg', '~> 1.2'
```

so applying the suggestion.


rafiss (approved):
stamp

PR #188, opened 2021-03-15
Call telemetry built-in at initialization time

solongordon:
I enhanced the initialize function to call the increment_feature_counter
built-in (if supported). This triggers telemetry on usage of the
adapter.

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb_adapter.rb
+++ b/lib/active_record/connection_adapters/cockroachdb_adapter.rb
@@ -209,6 +209,26 @@ module ActiveRecord
           version_num = 202
         end
         @crdb_version = version_num
+
+        if @config[:disable_cockroachdb_telemetry]
+          return
+        end
+        Thread.new do
+          pool.with_connection do |conn|
+            if !conn.active?
+              return
+            end
+            begin
+              query = "SELECT crdb_internal.increment_feature_counter('ActiveRecord %d.%d')"
+              conn.execute(query % [ActiveRecord::VERSION::MAJOR, ActiveRecord::VERSION::MINOR])
+            rescue ActiveRecord::StatementInvalid
+              # The increment_feature_counter built-in is not supported on this
+              # CockroachDB version. Ignore.
+            rescue => error
+              logger.warn "Unexpected error when incrementing feature counter: #{error}"
+            end
+          end
+        end
       end
 
       def self.database_exists?(config)


lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -205,6 +205,17 @@ def initialize(connection, logger, conn_params, config)
           version_num = 202
         end
         @crdb_version = version_num
+
+        begin
+          query_value("SELECT crdb_internal.increment_feature_counter('activerecord v6.1')")
```
solongordon, line 210: Can I get the Rails version number programatically here?

otan: /chime 

i am wondering if this creates an extra undesired round trip on each ORM. maybe this should be part of the connection parameters instead :\

solongordon: Good point, though I'm not sure that passing as a connection parameter would completely solve the problem since it would be rejected by earlier CRDB versions. You could then retry and omit the parameter, but then you haven't saved a round-trip.

solongordon: Hmm, though another nice thing about using a connection parameter is I think we could get away with not hashing the value before reporting. I'll think on this.

otan: Do we ignore or hard fail when we find a conn param we don't like?

Otherwise, I think most ORMs I think have a stanza where they send a bunch of initial sql commands up to pg catalog and the like, it'd be ideal if we tacked this on to one of them.

vy-ton:
> I enhanced the initialize function to call the increment_feature_counter
built-in (if supported). This triggers telemetry on usage of the
adapter.

Would this trigger once per tool? Or once per connection?

lib/active_record/connection_adapters/cockroachdb_adapter.rb
keithdoggett: Passing additional fields into `conn_params` won't raise an error.

It looks like those startup queries are run every time a connection pool is created through `establish_connection`, and they're made when we call the parent method here.

Also, `ActiveRecord::VERSION::STRING` will return the current version.

solongordon: CockroachDB will error out on an unrecognized connection parameter.

I updated my code to spawn a separate thread for this query since there's no reason to block on it, so hopefully this avoids any latency hit. Also incorporated `ActiveRecord::VERSION`, thanks for the pointer.

rafiss (approved):
lgtm! but i'll let keith weigh in about async patterns in ruby (in case there's anything to say)

rafiss:
hmm i wonder if this caused this error of if it's a flake

```
[17:18:22]
[Step 2/2] Error:
[17:18:22]
[Step 2/2] TestUnconnectedAdapter#test_error_message_when_connection_not_established:
[17:18:22]
[Step 2/2] 	from /home/agent/.gems/bundler/gems/rails-d203733d7872/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb:687:in `log'
[17:18:22]
[Step 2/2] ActiveRecord::ConnectionNotEstablished: connection is closed
```

lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -205,6 +205,20 @@ def initialize(connection, logger, conn_params, config)
           version_num = 202
         end
         @crdb_version = version_num
+
+        Thread.new do
+          begin
+            query = "SELECT crdb_internal.increment_feature_counter('ActiveRecord %d.%d')"
```
otan, line 211: does this use the same connection or a different one? my fear with async threads is that they cause surprises to the main thread such as closing a connection accidentally.

(very sad about the connection param deas)

otan: we should probably also gate this with an environment variable, e.g. `BLOCK_COCKROACHDB_ACTIVERECORD_TELEMETRY` which prevents spawning this thread if the user so wishes

rafiss: env var or some other configuration makes sense. i was asking @keithdoggett if we could also have configuration for using follower reads also, so whatever the config is, it should be a similar pattern for both

keithdoggett: I think something like #192 should work here. It would allow the user to set it in their `databases.yml` file and would be available in the `@config` hash here.

@otan to your point, it would be better to use `pool.with_connection` inside the thread because otherwise it will create zombie connections that may or may not be dealt with later. So it would look something like this

```ruby
Thread.new do
  pool.with_connection do |conn|
    begin
      query = "SELECT crdb_internal.increment_feature_counter('ActiveRecord %d.%d')"
      conn.query_value(query % [ActiveRecord::VERSION::MAJOR, ActiveRecord::VERSION::MINOR])
```

The connection is returned to the pool after the block finishes.

@rafiss I'm seeing multiple issues in that test class as well. I think it might be due to an error being raised in the thread and it's causing a different than expected error.

solongordon: Thanks for all the helpful feedback. I made the following changes:

- Used `pool.with_connection`
- Added check for `conn.active?`
- Added `disable_cockroachdb_telemetry` config variable
- Added another `rescue` clause to catch and log any unexpected exceptions since we never really want these to bubble up and cause more trouble.

Let's see if tests are happier now.

solongordon: Hmm, it looks like the `pool.with_connection` change causes this to be called three times when I run tests rather than just once. Perhaps because calling `with_connection` triggers another connection to be initialized? Any thoughts on how to avoid this @keithdoggett?

solongordon:
> Would this trigger once per tool? Or once per connection?

Sorry just saw this @vy-ton. It should be once per connection.

lib/active_record/connection_adapters/cockroachdb_adapter.rb
keithdoggett: Hmm, that's interesting. I thought that `with_connection` just checks out a connection from the pool, but maybe since we're calling this in `initialize`, no connections are in the pool yet and it creates one. I'll have some time to look into this more tomorrow.

keithdoggett: I looked into this a little bit and I think it's going to be hard to ensure that starting a new thread won't create a new connection within the `initialize` method. `with_connection` checks if there's a connection in the pool (on the first run there's not), if not it creates a new connection which ends up in `initialize` again and it tries to get a free connection from the pool or create a new one. So this will end up creating up to `MAX_THREAD_SIZE` connections or run until one becomes free and our new thread can just use that.

Do you want this to run once per connection or once every time we use the adapter? I wrote a little modification to the `ConnectionPool` that will run this in a new thread on initialization and only creates 1 connection, but it's only run when `ConnectionPool` is initialized.

If you want it to be run on every connection, one way we could go about it is to add some flag to the connection that indicates if this was run yet and keep track of the connections to wait until they're free to run them in a new thread. I'm not exactly sure how it would be implemented in a thread-safe manner since none of the thread-safe methods that `ConnectionPool` implements let you specify a connection.

solongordon: Once per use of the adapter would actually be preferable. If you'd like to share your `ConnectionPool` version that would be appreciated!

keithdoggett: https://github.com/keithdoggett/activerecord-cockroachdb-adapter/commit/a4c4580ec098bdd108cf2410490a7afd77eb0a11#. I ran the tests with this and there were 5 failures, which all seem to be due to the test trying to force only invalid connections to be available, but `with_connection` created a valid connection each time.
 
It might make sense to test this in a similar way to https://github.com/cockroachdb/activerecord-cockroachdb-adapter/commit/5f14dd422649ad9a68f594d3d555f194ed55825f where we change `config.yml`, `build/config.teamcity.yml`, and https://github.com/cockroachdb/activerecord-cockroachdb-adapter/commit/5f14dd422649ad9a68f594d3d555f194ed55825f#diff-8a06c7c920b8fb065a664471bffa5e1bdfca416b8eaea68ee2e0caec0026f8acR20 to set `disable_cockroachdb_telemetry` to `false` and add a new test to the `postgresql_adapter_test.rb` file that allows the thread to run.

Another option would be to establish a new connection before each of those tests that disables telemetry, then reset to the standard connection afterwards.

PR #189, opened 2021-03-22
Set supports_partitioned_indexes to false

solongordon:
This was false until we bumped our advertised PostgreSQL version in
v21.1. I set it back to false since it was causing a test failure.


rafiss (approved):
thanks for the fix!

Issue #48, opened 2020-04-02
[Tracking] Outstanding ActiveRecord 5.2 tests


alimi:
Below is a list of all the ActiveRecord tests that get run against PostgreSQL on Rails 5.2. Check off tests once they're passing against master.

There will be tests that don't work against CockroachDB. To skip these tests

1. Add a ruby file to `test/excludes` that matches the name of the test class if one doesn't already exist. For example to exclude a test from `ActiveRecord::AdapterTest`, create `test/excludes/ActiveRecord/AdapterTest.rb`.
1. Add an exclude statement to the file with the name of the test to exclude and a description. For example to exclude `test_indexes` from `ActiveRecord::AdapterTest`:
  https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/c3cb637bd855b36bbf9e76b7947162054b95022e/test/excludes/ActiveRecord/AdapterTest.rb#L1
1. Finally, if the test can run against CockroachDB with a few changes, add it to `test/cases`. Everything from the ActiveRecord test suite will be available, so a lot of the excluded test can be copied over. Namespace the test under the `CockroachDB` module to avoid name collisions. See [`test/cases/adapter_test.rb`](https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/c3cb637bd855b36bbf9e76b7947162054b95022e/test/cases/adapter_test.rb) for an example.

## ActiveRecord tests
- [ ] test/cases/adapter_test.rb
- [ ] test/cases/adapters/postgresql/active_schema_test.rb
- [ ] test/cases/adapters/postgresql/array_test.rb
- [ ] test/cases/adapters/postgresql/bit_string_test.rb
- [ ] test/cases/adapters/postgresql/bytea_test.rb
- [ ] test/cases/adapters/postgresql/case_insensitive_test.rb
- [ ] test/cases/adapters/postgresql/change_schema_test.rb
- [ ] test/cases/adapters/postgresql/cidr_test.rb
- [ ] test/cases/adapters/postgresql/citext_test.rb
- [ ] test/cases/adapters/postgresql/collation_test.rb
- [ ] test/cases/adapters/postgresql/composite_test.rb
- [ ] test/cases/adapters/postgresql/connection_test.rb
- [ ] test/cases/adapters/postgresql/datatype_test.rb
- [ ] test/cases/adapters/postgresql/date_test.rb
- [ ] test/cases/adapters/postgresql/domain_test.rb
[...]

PR #190, opened 2021-03-22
Set supports_partitioned_indexes to false

solongordon:
This was false until we bumped our advertised PostgreSQL version in
v21.1. I set it back to false since it was causing a test failure.


rafiss (changes requested):
looks like a different test skip was not backported here. could you skip `PostgresqlMoneyTest#test_money_regex_backtracking`? 

there are instructions here https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/48

solongordon:
I cherry-picked Oliver's skip commit, hopefully all good now.

rafiss (approved):
looks good! thanks

PR #192, opened 2021-03-28
Allow Follower Reads on pg_type Queries

keithdoggett:
Allows historical queries to be used on the two initialization queries that use the `pg_type` catalog. Passing the `use_follower_reads` option to the database config hash will configure which version of the query to use.

### DIFF
--- a/test/cases/adapters/postgresql/postgresql_adapter_test.rb
+++ b/test/cases/adapters/postgresql/postgresql_adapter_test.rb
@@ -15,6 +15,15 @@ module CockroachDB
         @connection_handler = ActiveRecord::Base.connection_handler
       end
 
+      def teardown
+        # use connection without follower_reads
+        database_config = { "adapter" => "cockroachdb", "database" => "activerecord_unittest" }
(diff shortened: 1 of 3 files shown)


test/cases/adapters/postgresql/postgresql_adapter_test.rb
```diff
@@ -28,6 +28,18 @@ def test_database_exists_returns_true_when_the_database_exists
         assert ActiveRecord::ConnectionAdapters::CockroachDBAdapter.database_exists?(db_config.configuration_hash),
           "expected database #{db_config.database} to exist"
       end
+
+      def test_using_follower_reads_connects_properly
+        database_config = { "use_follower_reads": true, "adapter" => "cockroachdb", "database" => "activerecord_unittest" }
```
rafiss, line 42: can this be called `use_follower_reads_for_type_introspection`? (or another name you like?) i want it to be more clear that it doesn't affect all queries

also, could you make a "configuration" section in the README that documents this setting, and any others we add later

rafiss (changes requested):
nice!! i just have a request for naming/documentation

rafiss (approved):
thank you!

PR #196, opened 2021-04-20
Call Telemetry on Connection Pool Initialization

keithdoggett:
Similar changes to #188, but instead of making the call in connection initialization, it is done during connection pool initialization. Differences in this PR are based on the discussion in #188. 

Notes:
* We decided to remove the call from the separate thread because it was causing applications to hang on startup. 
* I've added another condition to the guard statement before the telemetry call. It checks that the adapter is "cockroachdb". This is needed because the connection pool is adapter agnostic, so other adapters in the test suite would attempt to execute this statement. This would usually just result in a `StatementInvalid` exception or something harmless, but some tests use a `FakeAdapter` that don't actually implement `connection` and would cause the test suite to crash.

If these changes look good I can fix the merge conflict and add a note about `disable_cockroachdb_telemetry` to the README.

### DIFF
--- a/test/cases/adapters/postgresql/postgresql_adapter_test.rb
+++ b/test/cases/adapters/postgresql/postgresql_adapter_test.rb
@@ -15,6 +15,15 @@ module CockroachDB
         @connection_handler = ActiveRecord::Base.connection_handler
       end
 
+      def teardown
+        # use connection without follower_reads
+        database_config = { "adapter" => "cockroachdb", "database" => "activerecord_unittest" }
+        ar_config = ActiveRecord::Base.configurations.configs_for(env_name: "arunit", name: "primary")
+        database_config.update(ar_config.configuration_hash)
+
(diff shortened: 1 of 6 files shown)


test/cases/adapters/postgresql/postgresql_adapter_test.rb
```diff
@@ -15,6 +15,15 @@ def setup
         @connection_handler = ActiveRecord::Base.connection_handler
       end
 
+      def teardown
+        # use connection without follower_reads
```
solongordon, line 19: Could mention telemetry in this comment too.

solongordon (approved):
Looks good, thanks for the help!

test/cases/adapters/postgresql/postgresql_adapter_test.rb
keithdoggett: Done

PR #176, opened 2021-02-09
Add Activerecord 6.1 Support

keithdoggett:
Adds support for ActiveRecord 6.1. A lot of the changes are just adding new exclusions and re-implementations of tests locally, if applicable. There are a few notable additions though.

### Interval Type

In ActiveRecord 6.1, the interval type was added for Postgres. ActiveRecord makes the assumption that `IntervalStyle` can be set to `iso8601` in the database, but that is not possible in CockroachDB currently. This causes issues because the interval type assumes all interval string representations will be ISO8601 strings, but CockroachDB uses the standard Postgres format. To solve this an override of the interval type is added, as well as a Postgres interval parser and serializer.

Adds:
`lib/active_record/connection_adapters/cockroachdb/oid/interval.rb`
`test/cases/adapters/postgresql/interval_test.rb`
`test/cases/excludes/PostgresqlIntervalTest.rb`

### Adapter Changes

The `cockroachdb_adapter.rb` file has some significant changes.

* ConnectionHandling is updated to use the new connection creation style.
* `column_definitions` override is refactored and the `interval` and `numeric` types are added.
* `is_cached_plan_failure?` override is added because of the difference in how a `PreparedStatementCacheExpired` error is determined between Postgres and CockroachDB.
* The `numeric` type is overridden in `initialize_type_map` to properly check if `scale` was specified by the user.

### TemplateCreator

The `TemplateCreator` module was added. It allows a user to create/load the test schema from a template database, based on the ActiveRecord version number. The `CONTRIBUTING.md` doc was updated to show usage.

Adds:
`test/support/template_creator.rb`
`test/cases/helper.rb`: Duplicate of the one in ActiveRecord, but with guards around the `load_schema` call.

Issue #178, opened 2021-02-09
Numeric types have improper scale


keithdoggett:
In Postgres/CockroachDB if a column is defined with just precision and no scale, the scale is assumed to be 0 (ex. `NUMERIC(10)` -> `NUMERIC(10,0)`), but ActiveRecord treats this as a NUMERIC without scale defined, which means it can have digits to the right of the decimal place.

We are going to match the ActiveRecord behavior, but use this to track it and check if the behavior changes eventually.

PR #197, opened 2021-04-23
Fix Decimal type when precision and scale are unspecified

carlallen:
The changes to make intervals work broke decimal fields that do not specify precision and scale. https://github.com/cockroachdb/activerecord-cockroachdb-adapter/pull/176

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb_adapter.rb
+++ b/lib/active_record/connection_adapters/cockroachdb_adapter.rb
@@ -98,7 +98,7 @@ module ActiveRecord
           st_polygon:          {},
         }
 
-      # http://postgis.17.x6.nabble.com/Default-SRID-td5001115.html      
+      # http://postgis.17.x6.nabble.com/Default-SRID-td5001115.html
       DEFAULT_SRID = 0
 
       include CockroachDB::SchemaStatements
@@ -272,12 +272,9 @@ module ActiveRecord
             precision = extract_precision(sql_type)
             scale = extract_scale(sql_type)
 
-            # TODO(#178) this should never use DecimalWithoutScale since scale
-            # is assumed to be 0 if it is not explicitly defined.
-            #
             # If fmod is -1, that means that precision is defined but not
             # scale, or neither is defined.
-            if fmod && fmod == -1
+            if fmod && fmod == -1 && !precision.nil?
               # Below comment is from ActiveRecord
               # FIXME: Remove this class, and the second argument to
               # lookups on PG
@@ -390,7 +387,7 @@ module ActiveRecord
         # In general, it is hard to parse that, but it is easy to handle the common
         # case of an empty array.
         def extract_empty_array_from_default(default)
-          return unless supports_string_to_array_coercion? 
+          return unless supports_string_to_array_coercion?
           return unless default =~ /\AARRAY\[\]\z/
           return "{}"
         end
--- a/test/cases/adapters/postgresql/numeric_test.rb
+++ b/test/cases/adapters/postgresql/numeric_test.rb
@@ -1,0 +1,28 @@
+require "cases/helper_cockroachdb"
+
+# Load dependencies from ActiveRecord test suite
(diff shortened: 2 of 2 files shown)


lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -247,12 +247,9 @@ def initialize_type_map(m = type_map)
             precision = extract_precision(sql_type)
             scale = extract_scale(sql_type)
 
-            # TODO(#178) this should never use DecimalWithoutScale since scale
-            # is assumed to be 0 if it is not explicitly defined.
-            #
```
keithdoggett, line 252: I think this comment should still be here since ideally we wouldn't need this conditional and `OID::Decimal` would properly handle the case where precision is defined, but the scale is not.

test/cases/adapters/postgresql/numeric_test.rb
```diff
@@ -0,0 +1,32 @@
...
+    class PostgresqlNumber < ActiveRecord::Base; end
+
+  setup do
+    @connection = ActiveRecord::Base.connection
+    @connection.create_table("postgresql_numbers", force: true) do |t|
+      t.decimal 'decimal_default'
+      t.decimal 'decimal_with_p', precision: 5
+      t.decimal 'decimal_with_p_and_s', precision: 5, scale: 4
+    end
+  end
```
keithdoggett, line 19: Looks like the indentation's off here.

keithdoggett (commented):
Thanks for submitting this! The fix and tests look good, I just left a few notes for you to take a look at.

lib/active_record/connection_adapters/cockroachdb_adapter.rb
carlallen: Yeah, I removed it because it's not in the original rails code and the new logic is much closer to the original logic in rails.

keithdoggett: Yes, you're right. I think issue #178 can be closed since this handles the case described in that issue (`DECIMAL` where only precision is specified).

test/cases/adapters/postgresql/numeric_test.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    teardown do
+      @connection.drop_table "postgresql_decimals", if_exists: true
+    end
+
+    def test_decimal_values
+      record = PostgresqlNumber.new(decimal_default: 111.222, decimal_with_p: 111.222, decimal_with_p_and_s: 111.222)
+      assert_equal record.decimal_default, 111.222
+      assert_equal record.decimal_with_p, 11
```
keithdoggett, line 28: Looks like this is failing because the ActiveRecord Types don't handle precision properly. Both `Type::DecimalWithoutScale` and `OID::Decimal`  allow any number of digits to the left of the decimal point for some reason. I can look into it more to see if using the `limit` parameter helps.

carlallen: I can just drop that test since it's not as important, it's really only the first assert that I'm actually worried about here.

lib/active_record/connection_adapters/cockroachdb_adapter.rb
keithdoggett: Actually, can you add this back in now? I'm going to add to that issue to reflect the problem we saw with precision being improperly handled.

carlallen: I verified it does the same behavior in vanilla activerecord.

rafiss:
@keithdoggett thanks for reviewing and your explanation!

@carlallen would you mind rebasing on top of the latest changes from master and squashing this into one commit with a proper message? then i will merge this

carlallen:
Rebased and squashed with a more informative message

rafiss:
Thanks for finding this issue and fixing it @carlallen!  Just waiting for CI to pass, then I'll merge

PR #198, opened 2021-04-26
Update version to 6.1.0

solongordon:


rafiss (changes requested):
would you be able to add a line to https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/master/CHANGELOG.md

solongordon:
Yes, done!

rafiss (approved):

Issue #204, opened 2021-05-07
Version 6.1 with parallel_tests gem is not working anymore


Piioo:
Hello,

I tried to update the gem to the stable 6.1 version and our application is working fine, but we have problems now with the parallel_tests gem.
Im not sure, but it cant create databases anymore.

```
➜  foo git:(chore/update_activerecord-cockroachdb-adapter_non_beta) bundle exec rake parallel:setup
rake aborted!
ActiveRecord::NoDatabaseError: PG::InvalidCatalogName: ERROR:  database "foo_test5" does not exist
/Users/pisz/.rvm/gems/ruby-2.6.3/gems/activerecord-cockroachdb-adapter-6.1.0/lib/active_record/connection_adapters/cockroachdb_adapter.rb:49:in `rescue in cockroachdb_connection'
/Users/pisz/.rvm/gems/ruby-2.6.3/gems/activerecord-cockroachdb-adapter-6.1.0/lib/active_record/connection_adapters/cockroachdb_adapter.rb:27:in `cockroachdb_connection'
/Users/pisz/.rvm/gems/ruby-2.6.3/gems/activerecord-6.1.3.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:882:in `public_send'
/Users/pisz/.rvm/gems/ruby-2.6.3/gems/activerecord-6.1.3.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:882:in `new_connection'
/Users/pisz/.rvm/gems/ruby-2.6.3/gems/activerecord-6.1.3.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:926:in `checkout_new_connection'
/Users/pisz/.rvm/gems/ruby-2.6.3/gems/activerecord-6.1.3.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:905:in `try_to_checkout_new_connection'
/Users/pisz/.rvm/gems/ruby-2.6.3/gems/activerecord-6.1.3.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:866:in `acquire_connection'
/Users/pisz/.rvm/gems/ruby-2.6.3/gems/activerecord-6.1.3.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:588:in `checkout'
[...]

PR #205, opened 2021-05-13
Do not throw exception if the db doesn't exist

carlallen:
`rake db:create` is failing unless `disable_cockroachdb_telemetry` is explicitly set to `true`

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb_adapter.rb
+++ b/lib/active_record/connection_adapters/cockroachdb_adapter.rb
@@ -76,6 +76,8 @@ module ActiveRecord
             end
           end
         end
+      rescue ActiveRecord::NoDatabaseError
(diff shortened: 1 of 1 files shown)


keithdoggett:
Same as the PR I just opened (#206). We can just use this since it was first, but I also want to note that this fixes #204 as well as creation issues.

keithdoggett:
Although it may be better to wrap just the `with_connection` block with `begin/rescue` since this version will rescue from upstream errors `NoDatabaseErrors` in `super`. Although it's unlikely that `super` would ever generate that error in future versions it is probably safer to limit the scope as much as possible.

carlallen:
Yeah, your change is probably more correct, I'm cool with either one as long as one of them gets merged reasonably quickly.

Issue #169, opened 2021-01-11
Use built-in function before releasing 5.x with spatial support and 6.x GA release


vy-ton:
After https://github.com/cockroachdb/cockroach/issues/57480 is available, the function should be called from within the Active Record adapter code. Ideally, the arbitrary string should detect

- use of v5.x adapter
- use of v6.x adapter
- use of RGeo/PostGIS functionality (is there somewhere this could hook in?)

sandstrom:
I think the blocker is gone now:
https://github.com/cockroachdb/cockroach/pull/59768

rafiss:
@solongordon will look into doing this; or else delegate to Keith if needed.

vy-ton:
@solongordon Do you plan on backporting https://github.com/cockroachdb/activerecord-cockroachdb-adapter/pull/196 to 5.2 as well?

PR #210, opened 2021-06-06
5.2.x: Add telemetry call upon initialization

rafiss:
fixes https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/169

Backports commits:
- Call telemetry built-in at initialization time
- move telemetry thread to pool initialize
- Add adapter name guard before telemetry call and make call in the current thread

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb_adapter.rb
+++ b/lib/active_record/connection_adapters/cockroachdb_adapter.rb
@@ -53,18 +53,23 @@ module ActiveRecord
         adapter = pool_config.config[:adapter]
         return if disable_telemetry || adapter != "cockroachdb"
 
-        with_connection do |conn|
-          if conn.active?
-            begin
-              query = "SELECT crdb_internal.increment_feature_counter('ActiveRecord %d.%d')"
-              conn.execute(query % [ActiveRecord::VERSION::MAJOR, ActiveRecord::VERSION::MINOR])
-            rescue ActiveRecord::StatementInvalid
-              # The increment_feature_counter built-in is not supported on this
-              # CockroachDB version. Ignore.
-            rescue StandardError => e
-              conn.logger.warn "Unexpected error when incrementing feature counter: #{e}"
+
+        begin
+          with_connection do |conn|
+            if conn.active?
+              begin
+                query = "SELECT crdb_internal.increment_feature_counter('ActiveRecord %d.%d')"
+                conn.execute(query % [ActiveRecord::VERSION::MAJOR, ActiveRecord::VERSION::MINOR])
+              rescue ActiveRecord::StatementInvalid
+                # The increment_feature_counter built-in is not supported on this
+                # CockroachDB version. Ignore.
+              rescue StandardError => e
+                conn.logger.warn "Unexpected error when incrementing feature counter: #{e}"
+              end
             end
           end
+        rescue ActiveRecord::NoDatabaseError
+          # Prevent failures on db creation and parallel testing.
         end
       end
     end
--- a/lib/active_record/connection_adapters/cockroachdb_adapter.rb
+++ b/lib/active_record/connection_adapters/cockroachdb_adapter.rb
@@ -40,6 +40,14 @@ module ActiveRecord
       # PG::Connection object, so just pass a nil connection object for the
       # time being.
       ConnectionAdapters::CockroachDBAdapter.new(nil, logger, conn_params, config)
+    # This rescue flow appears in new_client, but it is needed here as well
+    # since Cockroach will sometimes not raise until a query is made.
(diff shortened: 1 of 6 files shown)


rafiss:
cc @keithdoggett -- could you review this whenever you have the time?

lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -46,6 +46,30 @@ def cockroachdb_connection(config)
...
+            begin
+              query = "SELECT crdb_internal.increment_feature_counter('ActiveRecord %d.%d')"
+              conn.execute(query % [ActiveRecord::VERSION::MAJOR, ActiveRecord::VERSION::MINOR])
+            rescue ActiveRecord::StatementInvalid
+              # The increment_feature_counter built-in is not supported on this
+              # CockroachDB version. Ignore.
+            rescue StandardError => e
+              conn.logger.warn "Unexpected error when incrementing feature counter: #{e}"
+            end
+          end
```
keithdoggett, line 67: We wrapped this in a `begin/rescue` block for version 6.x. We should probably do that here as well.

keithdoggett (commented):
Other than adding the `rescue` block I think this looks good. The big thing is reverting back from the `db_config` objects to hash-based configs, but it looks like you did that everywhere.

lib/active_record/connection_adapters/cockroachdb_adapter.rb
rafiss: isn't that there already? line 64 rescues all StandardErrors, which seems to be what 6.x is doing

keithdoggett: We found out that `with_connection` would raise errors if the target database didn't exist and had to wrap that block with `rescue` as well (#206).

test/cases/adapters/postgresql/postgresql_adapter_test.rb
```diff
@@ -0,0 +1,71 @@
...
+                        raise
+                      end
+                    end
+        assert_not db_exists, "expected database #{config[:database]} to not exist"
+      end
+
+      def test_database_exists_returns_true_when_the_database_exists
+        db_config = ActiveRecord::Base.configurations["arunit"]
+        db_exists = begin
+                      !!ActiveRecord::Base.postgresql_connection(config)
```
keithdoggett, line 46: Looks like `config` is undefined in this context (or calling another method). Should it be `!!ActiveRecord::Base.postgresql_connection(db_config)` instead?

keithdoggett:
We may also have to add a rescue statement at the end of the `cockroachdb_connection` method. https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/80c19c10229466db09bda6f6575a5f7877ddf255/lib/active_record/connection_adapters/cockroachdb_adapter.rb#L39-L54.

The `NoDatabaseError` is parsed as a `StatementInvalid` error, so this extra check was added to see if it was a `NoDatabaseError`.

lib/active_record/connection_adapters/cockroachdb_adapter.rb
rafiss: ah, i see. i missed that PR since i hadn't fetched the latest changes

test/cases/adapters/postgresql/postgresql_adapter_test.rb
rafiss: yes, nice catch

test/cases/adapters/postgresql/postgresql_adapter_test.rb
```diff
@@ -0,0 +1,71 @@
...
+        ActiveRecord::Base.establish_connection(database_config)
+      end
+
+      def test_database_exists_returns_false_when_the_database_does_not_exist
+        db_config = ActiveRecord::Base.configurations["arunit"]
+        config = db_config.dup
+        config[:database] = "non_extant_database"
+        db_exists = begin
+                      !!ActiveRecord::Base.postgresql_connection(config)
+                    rescue ::PG::Error => error
```
keithdoggett, line 33: I think we may have to rescue from `ActiveRecord::NoDatabaseError` or `ActiveRecord::ActiveRecordError` here since the `PG::Error` is usually nested inside as the `cause`.

Also, we'll have to modify the message on line 47 of `cockroachdb_adapter` to include the error message for the test to pass. Probably something like `raise ActiveRecord::NoDatabaseError, error.cause.message`.

rafiss:
@keithdoggett this should be ready for review finally! thanks for all your help

keithdoggett (approved):
LGTM!

PR #215, opened 2021-07-20
Add Additional Telemetry

keithdoggett:
* Programmatically define `COCKROACH_DB_ADAPTER_VERSION` in `lib/version.rb`.
* Add gem version telemetry.
* Use `conn.quote` to sanitize inputs to the telemetry calls.

### DIFF
--- a/activerecord-cockroachdb-adapter.gemspec
+++ b/activerecord-cockroachdb-adapter.gemspec
@@ -4,7 +4,7 @@ lib = File.expand_path('lib', __dir__)
 $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 
 require './lib/version.rb'
-version = ActiveRecord::CockroachDBAdapterVersion
+version = ActiveRecord::COCKROACH_DB_ADAPTER_VERSION
 
 Gem::Specification.new do |spec|
   spec.name          = "activerecord-cockroachdb-adapter"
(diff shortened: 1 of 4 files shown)


activerecord-cockroachdb-adapter.gemspec
```diff
@@ -1,10 +1,16 @@
...
-lib = File.expand_path('../lib', __FILE__)
+
+lib = File.expand_path('lib', __dir__)
 $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 
+version_line = File.open('./lib/active_record/connection_adapters/cockroachdb/version.rb') do |f|
+  f.find {|line| line.include?("VERSION")}
+end
+version = version_line.split('=').last.strip.gsub("\"", '')
+
```
keithdoggett, line 10: Ideally, we'd be able to require the module defined in `version.rb` or `load` it, but doing either of those breaks the test runner. I tried to look into it, but couldn't figure out why loading that module caused issues. Basically, all the subdirectories of `active_record` were not loaded in the case we tried to load in the version.

Assuming the structure of the file version file doesn't change, this workaround should suffice, but we could try to investigate it more later if we think it'll be an issue.

keithdoggett: We could also hardcode it and remember to change it in both places.

keithdoggett: For whatever reason including the version under the `ConnectionAdapters` module caused loading issues, so I just moved the constant to `ActiveRecord::CockroachDBAdapterVersion` and that fixed the issue.

rafiss (approved):
looks good to me!

i believe i've asked you this before, but remind me: does this telemetry happen once-per-connection-pool-init or once-per-connection init?

keithdoggett:
Once per connection pool init

Issue #223, opened 2021-08-26
Model.new doesn't set default values


FX-HAO:
Rails version: 5.2.6
activerecord-cockroachdb-adapter: 5.2.3
CRDB: v21.1.7 (x86_64-unknown-linux-gnu, built 2021/08/09 17:55:28, go1.15.14)

`Model.new` doesn't set default values. 
And `Model.create` creates a record without setting default values. we have to call `reload` to retrieve default values.

migration:
```
create_table :orders do |t|
  t.decimal :amount, precision: 32, scale: 16, default: 0
end
```

model:
```
class Order < ApplicationRecord
end
```

open a rails console
```
order = Order.new
puts order.amount # returned nil, should be 0.0

order = Order.create
puts order.amount # returned nil, should be 0.0
puts order.reload.amount # returned 0.0
```

rafiss:
Thanks for reporting and finding this!

If you'd like to make a PR, that'd be great. Otherwise we'll look at this closer later.

cc @keithdoggett

PR #224, opened 2021-08-29
handle scientific notation default values

FX-HAO:
https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/223

I couldn't find any test cases for `extract_value_from_default`. so if any test cases are needed please let me know.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -4,7 +4,7 @@
 FROM cockroachdb/example-orms-builder:20200413-1918
 
 # Native dependencies for libxml-ruby and sqlite3.
-RUN apt-get update -y && apt-get install -y \
+RUN apt-get --allow-releaseinfo-change update -y && apt-get install -y \
   libxslt-dev \
   libxml2-dev \
   libsqlite3-dev \
--- a/lib/active_record/connection_adapters/cockroachdb_adapter.rb
+++ b/lib/active_record/connection_adapters/cockroachdb_adapter.rb
@@ -352,7 +352,8 @@ module ActiveRecord
           super ||
             extract_escaped_string_from_default(default) ||
             extract_time_from_default(default) ||
-            extract_empty_array_from_default(default)
+            extract_empty_array_from_default(default) ||
+            extract_decimal_from_default(default)
         end
 
         # Both PostgreSQL and CockroachDB use C-style string escapes under the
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -8,3 +8,4 @@
 /spec/reports/
 /tmp/
 *.gem
+.idea
```
keithdoggett, line 11: Maybe keep the platform specific ignore just changed locally?

keithdoggett:
Thanks @FX-HAO this looks good! It would be good to add a test either here https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/master/test/cases/adapters/postgresql/numeric_test.rb or https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/master/test/cases/defaults_test.rb.

It probably belongs in the defaults_test, but it may be difficult to modify the setup depending on how complex it is in the original ActiveRecord test.

keithdoggett:
This is the original test file https://github.com/rails/rails/blob/6-1-stable/activerecord/test/cases/defaults_test.rb. It looks like we could add to this section in our version https://github.com/rails/rails/blob/6-1-stable/activerecord/test/cases/defaults_test.rb.

FX-HAO:
added tests, but don't know why the CI failed.

keithdoggett:
Thanks for doing that. @rafiss looks like the docker image might be outdated?

rafiss:
I'll take a look at the docker / CI issue

rafiss:
seems like it might be this https://stackoverflow.com/questions/68802802/repository-http-security-debian-org-debian-security-buster-updates-inrelease

FX-HAO:
is there any progress?

keithdoggett:
Hey @FX-HAO been away for a bit and haven't had much time to look more into this, but based on the link above it seems like changing the 

```
RUN apt-get update -y && apt-get install -y \
  libxslt-dev \
  libxml2-dev \
  libsqlite3-dev \
  && rm -rf /var/lib/apt/lists/*
```

line in `build/Dockerfile` to

```
RUN apt-get --allow-releaseinfo-change update -y && apt-get install -y \
  libxslt-dev \
  libxml2-dev \
  libsqlite3-dev \
  && rm -rf /var/lib/apt/lists/*
```

should work.

keithdoggett (approved):

keithdoggett:
LGTM! @rafiss you got anything else to add or am I good to merge?

rafiss:
this lgtm! thanks for the fix

though i was surprised to see that we are re-building the docker image each time. i'll work on a change to push the image and pull it when running CI

PR #226, opened 2021-09-22
Override SchemaStatements#foreign_keys

keithdoggett:
Override the `foreign_keys` method in `SchemaStatements` to help speed up the `disable_referential_integrity` method (https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/master/lib/active_record/connection_adapters/cockroachdb/referential_integrity.rb) which is causing slow tests.

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
@@ -32,6 +32,41 @@ module ActiveRecord
           end
         end
 
+        # override
+        # Modified version of the postgresql foreign_keys method.
+        # Replaces t2.oid::regclass::text with t2.relname since this is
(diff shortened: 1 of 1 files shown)


rafiss (approved):
lgtm! i just re-ran CI to see if that helps

keithdoggett:
Looks like it hit a foreign key issue again. Not sure exactly what is causing that. Maybe the order the foreign keys are added back in `disable_referential_integrity`?

rafiss:
Hm, that's a good theory...It does seem like `add_foreign_keys` appeared in the error trace

rafiss:
@keithdoggett i think it still is worth getting this in

keithdoggett:
Sounds good I'll merge once the CI finishes

PR #227, opened 2021-09-28
ActiveRecord 7

keithdoggett:
Working branch for ActiveRecord 7.

**Current ActiveRecord Version**: 7.0.0.alpha2

So far I've just made changes so that the tests run. Once we have a test run done, I can go through and resolve the issues from there.

### DIFF
--- a/activerecord-cockroachdb-adapter.gemspec
+++ b/activerecord-cockroachdb-adapter.gemspec
@@ -17,9 +17,11 @@ Gem::Specification.new do |spec|
   spec.description   = "Allows the use of CockroachDB as a backend for ActiveRecord and Rails apps."
   spec.homepage      = "https://github.com/cockroachdb/activerecord-cockroachdb-adapter"
 
-  spec.add_dependency "activerecord", "~> 6.1"
+  spec.add_dependency "activerecord", "~> 7.0.0.alpha2"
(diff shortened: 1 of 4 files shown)


keithdoggett:
It seems that a lot of the foreign key issues are related to the ones we've seen before.

This issue (https://github.com/cockroachdb/cockroach/issues/71496) should investigate the issue and hopefully give us an idea of some workarounds.

keithdoggett:
I decided to skip a few of the new fixture tests for now because they pass only if the ActiveRecord fixture tests are run first. I spent quite a while trying to figure out why but couldn't come up with a solution to reliably fix it. There are a few other tests that are flakey in the CI and fail depending on the order tests are run that I can't reproduce locally.

ActiveRecord 7 was released today though, so I bumped our gemspec to include that and hopefully, no new failures pop up.

keithdoggett:
The only failures on the last run were because of a few timeouts so I bumped the threshold from 3 to 5 minutes. I'm pretty sure this is due to those tests having lots of fixtures and our version of `disable_referential_integrity` being much slower than the implementation in the Postgres adapter.

jayzalowitz:
Just wanted to pop in here and offer any spare engineering help yall could need as this is blocking my company from launching.

PR #229, opened 2021-10-13
Cockroachdb v21.2 Support

keithdoggett:
Add support for CockroachDB v21.2.

**current version:** 21.2.0

Known issues:

* `PostgreSQLAdapterTest#test_database_exists_returns_false_when_database_does_not_exist`: (https://github.com/cockroachdb/cockroach/pull/71440) Will be resolved once tracked issue is fixed.
* `CheckConstraintsTest`: 2 tests are failing because removing or altering a constraint within a transaction does not work anymore. Should have an issue soon to track (https://github.com/cockroachdb/cockroach/issues/19444).
* Expression Indexes not formed properly as `IndexDefinition` objects. (https://github.com/cockroachdb/cockroach/issues/72041)

TODOs (so far):

* Add version identification in adapter for v21.2.
* Add interval style statement back to adapter


keithdoggett:
@rafiss do you know if any of the fixes mentioned in the first comment will be merged into the 21.2 release?

keithdoggett:
@rafiss do you have any update on those fixes will be included in this release or will we need to come up with workarounds for this version?

rafiss:
Sorry I missed this. https://github.com/cockroachdb/cockroach/pull/71440 and https://github.com/cockroachdb/cockroach/issues/72041 will both be available in v21.2.1 (not in v21.2.0)

keithdoggett:
Ok sounds good.

* I can add an additional check for this version to determine if the database doesn't exist.
* I can replicate the constraints tests within the Cockroach adapter and disable transactions for the failing tests.
* For the expression indexes we have a few options:
  * Disable expression indexes until 21.2.1 is released
  * Skip the failing test and just note that it won't work until 21.2.1
  * Try to override the column definition method to identify expression indexes where indkey isn't 0 then revert back to the upstream version once 21.2.1 is released.

I can go with whatever you think is best for the third issue.

rafiss:
"Disable expression indexes until 21.2.1 is released" sounds like the best option to me. Is there a way we can add a version gate to only allow the feature/test if the version is >= v21.2.1?

keithdoggett:
Yes, I added a sig fig to the crdb version attribute so we can account for patches.

keithdoggett:
@rafiss I added a helper to retry tests that fail because of the random foreign key error. This should help development until the root cause gets fixed.

rafiss:
The latest failure is from https://github.com/cockroachdb/cockroach/issues/71652 so retrying CI again

Also, unfortunately there was a change in our release schedule, so the expression indexes fix won't go out until 21.2.2

rafiss (approved):
lgtm! feel free to merge when you'd like

Issue #234, opened 2021-10-27
Ruby schema definitions download performance issues


daniel-crlabs:
This issue is regarding performance problems for the CRDB driver for Ruby, `Activerecord`. According to the customer,  this is not a break/fix issue necessarily or a bug, it is a performance type of concern.

# Issue

Rails pulls data types from schema definitions in CRDB. These queries are particularly slow in CockroachDB.
The theory is that these are queries on internal table that are poorly optimized because there is no index on the internal table. The latency is about 200ms (or more) per query, which runs once per connection.

# Workaround/Solution 

Rails allows you to cache the schema definitions, that way you don't have to pull the schema from CRDB.
This is the command to generate the cache: `bundle exec rake db:schema:cache:dump`

# Additional Technical Details

Under the covers, Rails pulls the schema definitions from `pg_attribute` and uses that to do the data type conversion in the ORM in rails.

CRDB driver code references:
[line 456, this code block this part is slow](https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/ca16b6a339936b0fb337fd36281d14ac1f9e0a13/lib/active_record/connection_adapters/cockroachdb_adapter.rb#L456)
[line 916, this code block this part is even slower](https://github.com/rails/rails/blob/766b84dfe14ea87c42e6fbf34b8acaafbac9cc39/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L916)

# Reproduction

I was able to replicate this behavior in my lab. It seems the initial correlation here is, the more records the table has, the longer it'll take for the query to run ( I did not notice this when the table was small/had less records -- less than 100 thousand). However, I also noticed this was not always the case, sometimes I also got a longer response time when the table had less records (around the 400 thousand threshold).
I also noticed the response time for the queries is sometimes not even always close to the same. I was able to run the same exact query and get a response within 107ms, and a few seconds later, the same query returned within 807ms and then again 185ms, 65ms , 124ms, etc ....
This makes it seem like it is being cached somehow, or dependent on some other factor, that causes it to be slower or faster.
The same observations above were found when I ran the 2nd query, the response times varied from 313ms, 99ms, 66ms, 15ms, 117ms. This also seems random, so I'm not sure what the actual cause might be.

1st query, this is the one that took the longest to complete:

```
root@lab-kub01:30007/test> SELECT a.attname, format_type(a.atttypid, a.atttypmod),
[...]

PR #235, opened 2021-11-04
Fetch comment and crdb_sql_type in single query for table

carlallen:
Solves #234

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb_adapter.rb
+++ b/lib/active_record/connection_adapters/cockroachdb_adapter.rb
@@ -425,7 +425,21 @@ module ActiveRecord
         #
         # @see: https://github.com/rails/rails/blob/8695b028261bdd244e254993255c6641bdbc17a5/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L829
         def column_definitions(table_name)
-          fields = super
+          fields = query(<<~SQL, "SCHEMA")
+              SELECT a.attname, format_type(a.atttypid, a.atttypmod),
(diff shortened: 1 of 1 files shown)


keithdoggett:
Looks like comments aren't being picked up correctly anymore based on the output from the CI. Also, it seems like there may be some differences in the output that are causing intervals and geographic types to be parsed improperly. You might have to `downcase` the type to fix that. Not sure what's going on with the comments right now.

keithdoggett (approved):
LGTM thanks for the work on this

lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -435,32 +449,33 @@ def column_definitions(table_name)
             /interval/,
             /numeric/
           ]
+
           re = Regexp.union(target_types)
```
rafiss, line 453: is `target_types` and this `regexp` unused now?

rafiss (changes requested):

lib/active_record/connection_adapters/cockroachdb_adapter.rb
carlallen: Yes it's in the condition on https://github.com/cockroachdb/activerecord-cockroachdb-adapter/pull/235/files#diff-aa6cbad03effcf32e5ea1d8867d8727eb6a605525828574bd3b2ef8c249f8f54R456

rafiss: missed that, thanks!

rafiss (approved):
lgtm!

PR #240, opened 2022-02-02
handle atttypmod being correct for decimals from v22.1

otan:


lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -291,21 +297,23 @@ def initialize_type_map(m = type_map)
...
-            # If fmod is -1, that means that precision is defined but not
-            # scale, or neither is defined.
-            if fmod && fmod == -1 && !precision.nil?
-              # Below comment is from ActiveRecord
-              # FIXME: Remove this class, and the second argument to
-              # lookups on PG
-              Type::DecimalWithoutScale.new(precision: precision)
-            else
-              OID::Decimal.new(precision: precision, scale: scale)
+          if needs_numeric_override?
```
rafiss, line 302: it seems that this is getting called before `@crdb_version` is defined
```
20:20:30   /home/agent/activerecord-cockroachdb-adapter/lib/active_record/connection_adapters/cockroachdb_adapter.rb:210:in `needs_numeric_override?': undefined method `<' for nil:NilClass (NoMethodError)

20:20:30     from /home/agent/activerecord-cockroachdb-adapter/lib/active_record/connection_adapters/cockroachdb_adapter.rb:302:in `initialize_type_map'

20:20:30     from /home/agent/.gems/bundler/gems/rails-f90e6d14b9c8/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb:261:in `initialize'

20:20:30     from /home/agent/activerecord-cockroachdb-adapter/lib/active_record/connection_adapters/cockroachdb_adapter.rb:233:in `initialize'

20:20:30     from /home/agent/activerecord-cockroachdb-adapter/lib/active_record/connection_adapters/cockroachdb_adapter.rb:40:in `new'

20:20:30     from /home/agent/activerecord-cockroachdb-adapter/lib/active_record/connection_adapters/cockroachdb_adapter.rb:40:in `cockroachdb_connection'
```

rafiss (changes requested):

lib/active_record/connection_adapters/cockroachdb_adapter.rb
otan: not sure how to fix this then :\

otan: `initialize_type_map` is called in `super.initialize`, but we need to be initialized before we can issue the query to get the crdb version

rafiss: what if we change the condition below from
```
if fmod && fmod == -1 && !precision.nil?
```
to
```
if fmod && ((fmod == -1 && !precision.nil?) || (fmod - 4 & 0xffff).zero?)
```

ref: https://github.com/rails/rails/blob/87d4d0f4126f64d991d40a1827de50935ddfdbff/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L569

otan: also these kind of checks are super broken for mixed version clusters no? is this initalize method global?

otan: ah that works

rafiss: yeah good point about mixed version clusters :(

i believe this initialize happens once per connection

rafiss (approved):
ty!

PR #242, opened 2022-02-24
use SHOW CLUSTER SETTING version to determine version

otan:
This is to gate mixed-version features behind the actual version in
which something may be available.


rafiss (approved):
lgtm!

otan:
keeps failing the flaky txn test, retrying

Issue #244, opened 2022-03-01
PG::InsufficientPrivilege: ERROR:  only users with the admin role are allowed to show cluster setting 'version'


gnomus:
Hello,

after updating the Gem to the newest Version (6.1.6) the above Error Showed up in our Systems.
It seems that the Change in [this PR](https://github.com/cockroachdb/activerecord-cockroachdb-adapter/pull/242) is causing it.

For obvious Reasons the Database Users of our Apps don't have the 'admin' role and thus the Gem is not usable in the current state.

It would be awesome if you could revert that Change for now or figure out a different Solution.

greetings

Edit: Additional I would like to mention that this problem was not cought by our CI because the single-node-setup by default has a User with all permissions and roles and there seems to be no option to change that.

rafiss:
Sorry about that. We will have a fix available later today. In the meantime, using the previous version of the gem is a workaround.

rafiss:
Your comment about testing is valuable too. We also missed this problem since our CI tests with an admin. It's definitely a gap we should try to fix.

gnomus:
Thanks for the fast reply.
For now we rolled back to the previous Version of the Gem.
Fortunatly the issue showed up on our Staging and not Production :D

PR #245, opened 2022-03-01
Fix cluster version check

rafiss:
fixes https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/244

SHOW CLUSTER SETTING is only available to admins, so we need to get the
version in a different way. Unfortunately, only v22.1 has a good way for
non-admins to get this info.


otan (approved):
Jesus :( we really should stop using root for our ORM tests.

Issue #203, opened 2021-05-03
schema.rb contains rowid for join tables without id


Piioo:
Hello,

when we create a join table without id, then the schema.rb looks good.
```ruby
create_table :foo_bars, id: false do |t|
   t.uuid :foo_id
   t.uuid :bar_id
end
```

`create_table "foo_bars", id: false, force: :cascade do |t|`

But when we run another migration, then into the table definition in schema.rb is added a `rowid`.
`t.bigserial "rowid", null: false`

keithdoggett:
I tried this and was able to replicate it. I saw that the SQL generated was correct though and does not include anything about rowid. After looking into it I came across this https://www.cockroachlabs.com/docs/v20.2/create-table.html#create-a-table. Which states:

> In CockroachDB, every table requires a primary key. If one is not explicitly defined, a column called rowid of the type INT is added automatically as the primary key, with the unique_rowid() function used to ensure that new rows always default to unique rowid values.

Maybe @rafiss knows if it's possible to bypass this, but it seems like this is the expected behavior and passing `id: false` into the `create_table` statement won't do anything.

Piioo:
But the problem is, that the rails default tasks do not work correct with this.

Here how the database looks like after migration:
<img width="1024" alt="Bildschirmfoto 2021-05-18 um 15 14 07" src="https://user-images.githubusercontent.com/4640423/118657466-cda0d800-b7eb-11eb-9764-3f89ec29650a.png">

But when I remove the whole datebase and load the schema from the schema.rb (without removing the rowid)
`bundle exec rails db:drop db:create db:schema:laod`

It looks like this:
<img width="1024" alt="Bildschirmfoto 2021-05-18 um 15 16 20" src="https://user-images.githubusercontent.com/4640423/118657740-0640b180-b7ec-11eb-8f8c-9c8033b796fe.png">

keithdoggett:
Ok, so the problem is that it will create a new `rowid_#{n}` row every time the schema is reloaded into the database? I think the best option is to just ignore the `id` parameter then if it doesn't do anything anyways.

Piioo:
Version 6.1.1 is different. It creates this in schema.rb:

`create_table "foo_users", primary_key: "rowid_1", force: :cascade do |t|`

but the bigserial is till there:

```ruby
  create_table "foo_users", primary_key: "rowid_1", force: :cascade do |t|
    t.bigint "foo_id"
    t.bigint "user_id"
[...]

PR #247, opened 2022-04-26
Support hidden and virtual columns

rafiss:
fixes https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/203

This allows us to fix issue with implicit rowid column being created
incorrectly when reloading a schema.


lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -500,8 +503,18 @@ def column_definitions(table_name)
             dtype = field[1]
             field[1] = crdb_fields[field[0]][2].downcase if re.match(dtype)
             field[7] = crdb_fields[field[0]][1]&.gsub!(/^\'|\'?$/, '')
+            field[9] = true if crdb_fields[field[0]][3]
```
otan, line 506: this is so hacky, lol!

otan (approved):

lib/active_record/connection_adapters/cockroachdb_adapter.rb
rafiss: yeah idk, kinda surprising that there's no way to access the fields by name!

otan: just the whole `NULL as is_hidden` then overwriting it later is *chefs kiss*

sandstrom:
Awesome @rafiss 🎉 🙏🏻

PR #255, opened 2022-06-20
Fix broken and flaky tests

marmitoTH:
This pull request will fix all `Argument Error` and `Mock Exception Error`, as well as the performance tests that fail from time to time.


rafiss (approved):
thank you!

PR #262, opened 2023-03-23
retry ConnectionNotEstablished errors

ecwall:
I tried converting the monkey patch from this support ticket. Is this a proper application?

https://github.com/cockroachlabs/support/issues/2155#issuecomment-1475427677

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb/transaction_manager.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/transaction_manager.rb
@@ -10,20 +10,43 @@ module ActiveRecord
         # exceeded.
         def within_new_transaction(isolation: nil, joinable: true, attempts: 0)
           super(isolation: isolation, joinable: joinable)
+        rescue ActiveRecord::ConnectionNotEstablished => error
+          raise unless retryable? error
+          raise if attempts >= @connection.max_transaction_retries
+
+          sleep_rand_seconds(attempts)
+
+          unless @connection.active?
+            warn "connection isn't active, reconnecting"
+            @connection.reconnect!
+          end
+
+          within_new_transaction(isolation: isolation, joinable: joinable, attempts: attempts + 1) { yield }
         rescue ActiveRecord::StatementInvalid => error
           raise unless retryable? error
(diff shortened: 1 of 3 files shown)


lib/active_record/connection_adapters/cockroachdb/transaction_manager.rb
```diff
@@ -10,6 +10,20 @@ module TransactionManagerMonkeyPatch
         # exceeded.
         def within_new_transaction(isolation: nil, joinable: true, attempts: 0)
           super(isolation: isolation, joinable: joinable)
+        rescue ActiveRecord::ConnectionNotEstablished => e
+          raise unless retryable? error
+          raise if attempts >= @connection.max_transaction_retries
+
+          attempts += 1
+          sleep_seconds = (2 ** attempts + rand) / 10
```
chengxiong-ruan, line 18: let's make a method for this `sleep_seconds` and `sleep`, it's needed by the existing error handling below, maybe something like `sleep_rand_seconds`?

lib/active_record/connection_adapters/cockroachdb/transaction_manager.rb
```diff
@@ -10,21 +10,44 @@ module TransactionManagerMonkeyPatch
...
         end
+
+        def serialization_error?(error)
+          errors = [error]
+          errors << error.cause if error.cause
+          errors.any? {|e| e.is_a? PG::TRSerializationFailure }
+        end
+
+        def sleep_rand_seconds()
+          sleep_seconds = (2 ** attempts + rand) / 10
```
chengxiong-ruan, line 48: `attempts` should be a parameter of the method.

ecwall: Fixed.

lib/active_record/connection_adapters/cockroachdb/transaction_manager.rb
ecwall: Added.

ecwall:
bors r=chengxiong-ruan

PR #263, opened 2023-03-27
Use .ruby-version to control ruby version in local dev and ci

ecwall:


build/teamcity-test.sh
```diff
@@ -16,9 +24,6 @@ run_cockroach() {
   # Start a CockroachDB server, wait for it to become ready, and arrange
   # for it to be force-killed when the script exits.
   rm -f "$urlfile"
-  # Clean out a past CockroachDB instance. This will clean out leftovers
-  # from the build agent, and also between CockroachDB runs.
-  cockroach quit --insecure || true
```
ecwall, line 21: `cockroach quit` no longer exists

.ruby-version
```diff
@@ -0,0 +1 @@
+2.7.7
```
rafiss, line 1: is this actually what this project was using before? i wonder if we can get it onto 3.1.3

build/teamcity-test.sh
rafiss: do we need something else that shuts down the process?

rafiss (approved):

ecwall:
CI tests are failing, but they fail without this change also

PR #267, opened 2023-05-22
Fix default numbers test

andyyang890:
This patch fixes the default numbers test to expect the correct result after https://github.com/cockroachdb/cockroach/pull/102299 was merged.


rafiss (approved):
lgtm!

PR #270, opened 2023-06-21
Make tests runnable locally without configuration

BuonOmo:
See message per commit for details. This is the minimum amount of changes that lead me to successfully run the tests locally.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,58 +1,49 @@
-require 'openssl'
-source 'https://rubygems.org'
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
 gemspec
 
-if ENV['RAILS_SOURCE']
-  gemspec path: ENV['RAILS_SOURCE']
-else
-  def get_version_from_gemspec
-    gemspec = eval(File.read('activerecord-cockroachdb-adapter.gemspec'))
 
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -1,58 +1,49 @@
...
-      find { |major, minor|
-        major == gemspec_major && (minor == gemspec_minor || gemspec_minor.nil?)
-      }.join(".")
-  end
+    def gemspec_requirement
+      File
+        .foreach("activerecord-cockroachdb-adapter.gemspec", chomp: true)
+        .find { _1[/add_dependency\s.activerecord.,\s.(.*)./] }
+
+      Gem::Requirement.new(Regexp.last_match(1))
```
BuonOmo, line 20: Another option is to go as we did in the postgis adapter:

```suggestion
      gs = Bundler.load_gemspec("activerecord-postgis-adapter.gemspec")
      gs.dependencies.find { |d| d.name == "activerecord" }
```

and replace `satisfied_by?` with `match?` in the `.call` method. I prefer the current version though as we don't need to load gemspec for such little, and this regexp should be robust enough to last until the decay of ruby anyway

rafiss (approved):
very nice cleanup, thank you

PR #271, opened 2023-06-23
fix(column): avoid monkeypatching pg

### LINKED ISSUE #251: Incorrect db/schema.rb file for Postgres when using Cockroach and Postgres together.
In our Rails app we use both Postgres and Cockroach at the same time with Rails 6 multi-database support.

This adapter monkey patches `ActiveRecord::ConnectionAdapters::PostgreSQL::Column#serial?` in a way that causes the Postgres `db/schema.rb` file to be incorrectly generated, with primary key values like:

```
create_table "access_logs", id: :bigint, default: -> { "nextval('admin_access_logs_id_seq'::regclass)" }, force: :cascade do |t|
```

Instead of:

```
create_table "access_logs", force: :cascade do |t|
```

We've applied this patch below to fix the issue so that the `@serial` value is still respected when it's been supplied for Postgres columns.

```
module ActiveRecord
  # activerecord-cockroachdb-adapter monkey patches ActiveRecord::ConnectionAdapters::PostgreSQL::Column and
  # overwrites the #serial? method in a way that works for Cockroach, but always returns false for Postgres.
  #
  # This affects the generation of db/schema.rb for Postgres because the primary key column is not identified.
  #
  # Replace the method with one that works for both Postgres and Cockroach.
  #
  # I'm not sure why activerecord-cockroachdb-adapter monkey patches the PostgreSQL::Column class instead of extending
  # it, like it does for other classes.
  module FixDbSchemaDump
    # This combines two methods:
    #  - ActiveRecord::ConnectionAdapters::PostgreSQL::Column#serial?
    #  - ActiveRecord::ConnectionAdapters::CockroachDB::PostgreSQLColumnMonkeyPatch#serial?
    #
    # @serial will always be nil for Cockroach, but may be set for Postgres.
    def serial?
      @serial || default_function == "unique_rowid()"
    end
  end

  ConnectionAdapters::PostgreSQL::Column.class_eval do
    prepend FixDbSchemaDump
  end
end
```

What's the reason for the CockroachDB adapter monkey patching the PostgresSQL::Column class instead of creating a subclass?

BuonOmo:
Fixes #251

TODO:
- [ ] verify that it works
- [ ] implement tests

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb/column.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/column.rb
@@ -1,7 +1,7 @@
 module ActiveRecord
   module ConnectionAdapters
     module CockroachDB
-      module PostgreSQLColumnMonkeyPatch
+      class Column < PostgreSQLColumn
         # most functions taken from activerecord-postgis-adapter spatial_column
         # https://github.com/rgeo/activerecord-postgis-adapter/blob/master/lib/active_record/connection_adapters/postgis/spatial_column.rb
         def initialize(name, default, sql_type_metadata = nil, null = true,
@@ -92,10 +92,6 @@ module ActiveRecord
           end
         end
       end
-    end
-
-    class PostgreSQLColumn
-      prepend CockroachDB::PostgreSQLColumnMonkeyPatch
     end
   end
 end
--- a/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
@@ -90,7 +90,7 @@ module ActiveRecord
           # {:dimension=>2, :has_m=>false, :has_z=>false, :name=>"latlon", :srid=>0, :type=>"GEOMETRY"}
           spatial = spatial_column_info(table_name).get(column_name, type_metadata.sql_type)
 
-          PostgreSQL::Column.new(
+          CockroachDB::Column.new(
             column_name,
             default_value,
             type_metadata,
@@ -112,7 +112,7 @@ module ActiveRecord
         # since type alone is not enough to format the column.
         # Ex. type_to_sql(:geography, limit: "Point,4326")
         # => "geography(Point,4326)"
-        # 
+        #
         def type_to_sql(type, limit: nil, precision: nil, scale: nil, array: nil, **) # :nodoc:
           sql = \
             case type.to_s


rafiss (commented):
the changes seem fine to me. let's add a test to make sure the reported issue is resolved, and we don't regress.

BuonOmo:
@rafiss I've been looking at how we could test this. And it seems like we would need an active PG connection. I'm not sure it is worth it having a PG connection for a non-regression on this. I have already spent 30m on trying to find a simpler way to test this. I think I could manage a hack but I would need much more time, and I am not sure it is worth it.

I would still do some more manual testing before merging though

BuonOmo:
@rafiss so looking further, we also monkeypatched the method `adapter_name_from`. I think this PR could be changed to _support multiple databases connections_. WDYT?

rafiss:
I wanted to make sure I understand: Can you explain more about why making that change would support multiple database connections? And also could you define what you mean by "multiple database connections"?

BuonOmo:
> Can you explain more about why making that change would support multiple database connections?

Not sure which change you are talking about. But basically I want to make all of the sufficient changes to support _multiple database connections_. This may include more than the change made in this PR and the change mentioned in my above comment. For instance, I should un-exclude some tests and see what happen with those.

> And also could you define what you mean by "multiple database connections"?

Quoting https://guides.rubyonrails.org/active_record_multiple_databases.html:

> Rails now has support for multiple databases so you don't have to store your data all in one place.
> 
> At this time the following features are supported:
> 
> - Multiple writer databases and a replica for each
> - Automatic connection switching for the model you're working with
> - Automatic swapping between the writer and replica depending on the HTTP verb and recent writes
> - Rails tasks for creating, dropping, migrating, and interacting with the multiple databases

These database can even be set with various adapters. This is what the issue #251 is about!

---

Back on what I said previously:

> I'm not sure it is worth it having a PG connection for a non-regression on this. 

We could do some testing by shipping a sqlite connection, this is way lighter and could work. But it is still a great amount of work, and I don't know that you'd want to invest that much on those kind of tests having the main tests already flaky. tl;dr: for me testing this further is a _nice add_.

BuonOmo:
> Can you explain more about why making that change would support multiple database connections?

Not sure which change you are talking about. But basically I want to make all of the sufficient changes to support _multiple database connections_. This may include more than the change made in this PR and the change mentioned in my above comment. For instance, I should un-exclude some tests and see what happen with those.

> And also could you define what you mean by "multiple database connections"?

Quoting https://guides.rubyonrails.org/active_record_multiple_databases.html:

> Rails now has support for multiple databases so you don't have to store your data all in one place.
> 
> At this time the following features are supported:
> 
> - Multiple writer databases and a replica for each
> - Automatic connection switching for the model you're working with
> - Automatic swapping between the writer and replica depending on the HTTP verb and recent writes
> - Rails tasks for creating, dropping, migrating, and interacting with the multiple databases

These database can even be set with various adapters. This is what the issue #251 is about!

---

Back on what I said previously:

> I'm not sure it is worth it having a PG connection for a non-regression on this. 

We could do some testing by shipping a sqlite connection, this is way lighter and could work. But it is still a great amount of work, and I don't know that you'd want to invest that much on those kind of tests having the main tests already flaky. tl;dr: for me testing this further is a _nice add_.

PR #272, opened 2023-06-27
feat: add table related optimizer hints

### LINKED ISSUE #266: Add support for optimizer hints
**Is your feature request related to a problem? Please describe.**
The CRDB Rails adapter is not currently equipped to process the different format being used by the customer for hinting no_zigzag_join/force_zig_zag. Customer wants to minimize the amount of effort required to migrate from Aurora to CRDB.

**Describe the solution you'd like**
Integrate the PR created by the customer into our CRDB Rails adapter. This involves being able to force an index on a query through optimizer hints
https://www.cockroachlabs.com/docs/v22.2/table-expressions#force-index-selection

**Describe alternatives you've considered**
The customer is manually incorporating RAW SQL queries into the ActiveRecord adapter which is a labor-intensive

Current ticket open 
https://github.com/cockroachlabs/support/issues/2155#issuecomment-1485671733

Jira issue: CRDB-26266

BuonOmo:
TODO:
- [x] documentation
- [x] tests
- [x] refining API (edge cases)
- [ ] ~handle connections that do not have these hints available~ (won't do for now)

Fixes #266

### DIFF
--- a/lib/active_record/relation/query_methods_ext.rb
+++ b/lib/active_record/relation/query_methods_ext.rb
@@ -1,0 +1,98 @@
+# frozen_string_literal: true
+
+module ActiveRecord
+  class Relation
+    module QueryMethodsExt
+      def from!(...) # :nodoc:
+        @force_index = nil
+        @index_hint = nil
+        super
+      end
+
+      # Set table index hint for the query to the
+      # given `index_name`, and `direction` (either
+      # `ASC` or `DESC`).
+      #
+      # Any call to `ActiveRecord::QueryMethods#from`
+      # will reset the index hint. Index hints are
+      # not set if the `from` clause is not a table
+      # name.
+      #
+      # @see https://www.cockroachlabs.com/docs/v22.2/table-expressions#force-index-selection
+      def force_index(index_name, direction: nil)
+        spawn.force_index!(index_name, direction: direction)
+      end
+
+      def force_index!(index_name, direction: nil)
+        return self unless from_clause_is_a_table_name?
+
+        index_name = sanitize_sql(index_name.to_s)
+        direction = direction.to_s.upcase
+        direction = %w[ASC DESC].include?(direction) ? ",#{direction}" : ""
+
+        @force_index = "FORCE_INDEX=#{index_name}#{direction}"
+        self.from_clause = build_from_clause_with_hints
+        self
+      end
+
+      # Set table index hint for the query with the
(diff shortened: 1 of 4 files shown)


lib/active_record/relation/query_methods_ext.rb
```diff
@@ -0,0 +1,98 @@
...
+      # Set table index hint for the query to the
+      # given `index_name`, and `direction` (either
+      # `ASC` or `DESC`).
+      #
+      # Any call to `ActiveRecord::QueryMethods#from`
+      # will reset the index hint. Index hints are
+      # not set if the `from` clause is not a table
+      # name.
+      #
+      # @see https://www.cockroachlabs.com/docs/v22.2/table-expressions#force-index-selection
```
BuonOmo, line 21: I don't really like doing that as I saw that links to older versions of the doc are deleted. But I didn't find an alternative (for instance, `latest` instead of `v22.2`)

rafiss: good question! the info has moved here: https://www.cockroachlabs.com/docs/stable/select-clause.html#force-reverse-scan

rafiss (changes requested):
this is great! thanks for the good testing. i think we should document this in CHANGELOG.md. and possibly also in README.md, but i'm not sure if we need to.

i will mark it with "request changes" just to take care of the docs link and CHANGELOG update, but the code all looks good.

BuonOmo:
> and possibly also in README.md, but i'm not sure if we need to.

I didn't do it on purpose: I think this feature is cumbersome and I wouldn't bother users with it. If they need it they'll find it by looking in issues/pr, or [the doc](https://www.rubydoc.info/gems/activerecord-cockroachdb-adapter/). I can actually do some work on documentation to facilitate its development. I have done this in previous projects.

BuonOmo:
> That would be great! If you know a better way to organize the docs, that is definitely welcome.

Yes, you can have a look at the same page for RGeo (https://www.rubydoc.info/gems/rgeo/) for instance, all methods are listed. I'll open an issue on the topic to see how dikshant wants to prioritize!

> The doc that you linked to is generated from the README.md file of this project. So that's why I was thinking adding it there could make sense. But the CHANGELOG is good enough for now.

Not only the readme, have a look at the menu on the side, it is referencing the whole api. Though a bit poorly, hence the work I can do to make it better :) (mostly documenting better)

PR #273, opened 2023-07-04
feat: support sql dump

### LINKED ISSUE #2: Support db:structure:dump and db:structure:load
The Postgres DatabaseTasks implementation uses `pg_dump` to dump the database in `db:structure:dump`. CockroachDB doesn't support `pg_dump`, so this plugin should use `DUMP` and its inverse when they're officially supported.

BuonOmo:
Fixes #2

TODO:

- [x] support or explicitly fail for `ActiveRecord.dump_schemas` conf
- [x] explicitly fail having extra_flags
- [x] support for `load_schema` (see #275)
- [x] dump the content of the version table (by reading the code, it is already done and tested in AR, we're fine)

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb/database_tasks.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/database_tasks.rb
@@ -5,7 +5,43 @@ module ActiveRecord
     module CockroachDB
       class DatabaseTasks < ActiveRecord::Tasks::PostgreSQLDatabaseTasks
         def structure_dump(filename, extra_flags=nil)
-          raise "db:structure:dump is unimplemented. See https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/2"
+          if extra_flags
+            raise "No flag supported yet, please raise an issue if needed. " \
+              "https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/new"
+          end
+
+          case ActiveRecord.dump_schemas
+          when :all, String
+            raise  "Custom schemas are not supported in CockroachDB. " \
+              "See https://github.com/cockroachdb/cockroach/issues/26443."
+          when :schema_search_path
+            if configuration_hash[:schema_search_path]
+              raise  "Custom schemas are not supported in CockroachDB. " \
+                "See https://github.com/cockroachdb/cockroach/issues/26443."
+            end
+          end
+
+          conn = ActiveRecord::Base.connection
+          File.open(filename, "w") do |file|
+            %w(SCHEMAS TYPES).each do |object_kind|
+              ActiveRecord::Base.connection.execute("SHOW CREATE ALL #{object_kind}").each_row { file.puts _1 }
+            end
+
+            ignore_tables = ActiveRecord::SchemaDumper.ignore_tables.to_set
+
+            conn.execute("SHOW CREATE ALL TABLES").each_row do |(sql)|
+              if sql.start_with?("CREATE")
+                table_name = sql[/CREATE TABLE (?:.*?\.)?\"?(.*?)[\" ]/, 1]
+                next if ignore_tables.member?(table_name)
+              elsif sql.start_with?("ALTER")
+                table_name = sql[/ALTER TABLE (?:.*?\.)?\"?(.*?)[\" ]/, 1]
+                ref_table_name = sql[/REFERENCES (?:.*?\.)?\"?(.*?)[\" ]/, 1]
+                next if ignore_tables.member?(table_name) || ignore_tables.member?(ref_table_name)
+              end
+
+              file.puts sql
+            end
+          end
         end
 
         def structure_load(filename, extra_flags=nil)
--- a/test/cases/tasks/cockroachdb_rake_test.rb
+++ b/test/cases/tasks/cockroachdb_rake_test.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+
+require "cases/helper"
+require "active_record/tasks/database_tasks"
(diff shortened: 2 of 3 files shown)


lib/active_record/connection_adapters/cockroachdb/database_tasks.rb
```diff
@@ -5,7 +5,37 @@ module ConnectionAdapters
...
+            %w(SCHEMAS TYPES).each do |object_kind|
+              ActiveRecord::Base.connection.execute("SHOW CREATE ALL #{object_kind}").each_row { file.puts _1 }
+            end
+
+            ignore_tables = ActiveRecord::SchemaDumper.ignore_tables
+            all_tables = conn.execute("SHOW TABLES").field_values("table_name")
+            tables = all_tables - ignore_tables
+
+            tables.each do
+              create_stmt = conn.execute("SHOW CREATE TABLE #{conn.quote_table_name(_1)}").field_values("create_statement").first
```
BuonOmo, line 35: This implies an ugly amount of back and forth between the cockroach instance and the rails server. I'm looking into a better solution

BuonOmo:
@rafiss it is all yours!

I'm pretty sure there are some edge cases we are not covering as we are basically recreating `pg_dump` in 38 lines (the `structure_dump` method). Which is impossible for sure.

If you already see some of them, I'll implement them, otherwise I think that we should just wait and see issues coming.. I can issue a warning somewhere in the task saying that this behaviour might not dump the whole db if you want

lib/active_record/connection_adapters/cockroachdb/database_tasks.rb
```diff
@@ -5,7 +5,43 @@ module ConnectionAdapters
...
+          end
+
+          case ActiveRecord.dump_schemas
+          when :all, String
+            raise  "Custom schemas are not supported in CockroachDB. " \
+              "See https://github.com/cockroachdb/cockroach/issues/26443."
+          when :schema_search_path
+            if configuration_hash[:schema_search_path]
+              raise  "Custom schemas are not supported in CockroachDB. " \
+                "See https://github.com/cockroachdb/cockroach/issues/26443."
```
rafiss, line 20: custom schemas are well-supported in CockroachDB these days. can we support this case?

lib/active_record/connection_adapters/cockroachdb/database_tasks.rb
```diff
@@ -5,7 +5,43 @@ module ConnectionAdapters
...
+
+          conn = ActiveRecord::Base.connection
+          File.open(filename, "w") do |file|
+            %w(SCHEMAS TYPES).each do |object_kind|
+              ActiveRecord::Base.connection.execute("SHOW CREATE ALL #{object_kind}").each_row { file.puts _1 }
+            end
+
+            ignore_tables = ActiveRecord::SchemaDumper.ignore_tables.to_set
+
+            conn.execute("SHOW CREATE ALL TABLES").each_row do |(sql)|
```
rafiss, line 32: there are a few other statements that we will probably need in order to fully dump everything: `SHOW CREATE ALL TYPES` and `SHOW CREATE ALL SCHEMAS`

rafiss (changes requested):

lib/active_record/connection_adapters/cockroachdb/database_tasks.rb
BuonOmo: There are other places where we don't support it, so I think we should support it all at once, in a dedicated milestone

lib/active_record/connection_adapters/cockroachdb/database_tasks.rb
BuonOmo: We do, see lines 26 27 :)

rafiss: Thanks for pointing that out :)

lib/active_record/connection_adapters/cockroachdb/database_tasks.rb
rafiss: That sounds good to me!

test/cases/tasks/cockroachdb_rake_test.rb
```diff
@@ -0,0 +1,46 @@
...
+        returns: ["accounts", "articles"]
+      ) do
+        ActiveRecord::Tasks::DatabaseTasks.structure_dump(config, @filename)
+
+        read = File.read(@filename)
+      end
+
+      read = File.read(@filename)
+      refute read.include?("NOT TODAY"), "The dump file previous content was not overwritten"
+      assert read.include?("CREATE TABLE public.schema_migrations"), "No dump done"
```
rafiss, line 41: should the schema and type statements be tested here too?

rafiss (approved):
code looks good to me! i just had one testing question.

lib/active_record/connection_adapters/cockroachdb/database_tasks.rb
BuonOmo: #276

test/cases/tasks/cockroachdb_rake_test.rb
BuonOmo: well that is a good question. The tests in the ar test suit are even simpler than that (just checking for a comment). I can add those as they don't cost much to add. But I'm not sure of how secure this is. As stated before, I think an external and complete dump tool would be necessary to ensure some robustness!

rafiss: very fair point. i think the basic check you've added will at least catch a few regressions, even if it doesn't test each edge case.

rafiss (approved):

PR #275, opened 2023-07-06
feat: support sql load

BuonOmo:
NOTE: needs to be rebased on top of the sql-dump branch (not available as base branch since in other repo)

TODO:
- [x] tests
  - with already existing schema / table (also we need to decide of the behaviour in that case)
- [x] check that the insecure flag doesn't mess with secure params

On insecure, @rafiss I think there is a seemingly inconsistent behaviour in the cockroach client:

```bash
# This would work, prioritizing insecure mode
COCKROACH_URL='postgres://localhost:26257?sslmode=require' cockroach sql --insecure

# This wouldn't, prioritizing secure mode
cockroach sql --insecure --url='postgres://localhost:26257?sslmode=require'

# works
cockroach sql --url='postgres://localhost:26257?sslmode=require' --insecure

# doesn't work
COCKROACH_URL='postgres://localhost:26257?sslmode=require' COCKROACH_INSECURE=true cockroach sql

# doesn't work (order doesn't matter with env)
COCKROACH_INSECURE=true COCKROACH_URL='postgres://localhost:26257?sslmode=require' cockroach sql

```

My two cents is that the priority should always be given to one or the other, or that this behaviour should be documented.

### DIFF
--- a/test/support/rake_helpers.rb
+++ b/test/support/rake_helpers.rb
@@ -17,8 +17,7 @@ module RakeHelpers
     ar_test_files = ar_test_files.
       split(',').
       map { |file| File.join ARTest::CockroachDB.root_activerecord, file.strip }.
-      then { _1.prepend(COCKROACHDB_TEST_HELPER) unless _1.empty? }.
-      prepend(COCKROACHDB_TEST_HELPER)
+      tap { _1.prepend(COCKROACHDB_TEST_HELPER) unless _1.empty? }
 
     cr_test_files = cr_test_files.split(',').map(&:strip)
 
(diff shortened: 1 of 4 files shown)


test/support/rake_helpers.rb
```diff
@@ -17,8 +17,7 @@ def test_files
     ar_test_files = ar_test_files.
       split(',').
       map { |file| File.join ARTest::CockroachDB.root_activerecord, file.strip }.
-      then { _1.prepend(COCKROACHDB_TEST_HELPER) unless _1.empty? }.
-      prepend(COCKROACHDB_TEST_HELPER)
+      tap { _1.prepend(COCKROACHDB_TEST_HELPER) unless _1.empty? }
```
BuonOmo, line 20: I forgot that `Kernel#then` returns the result of the block, hence when the condition was false it would return nul rather than an empty array. `Kernel#tap` is what we want: it returns the value of `self`, here the modified array.

rafiss:
In general, the way to think about this the CLI is that the environment variables are setting the defaults, and the `--insecure` and `--url` flags are explicitly choosing which behavior.

As far as this example:
```
# This wouldn't, prioritizing secure mode
cockroach sql --insecure --url='postgres://localhost:26257?sslmode=require'

# works
cockroach sql --url='postgres://localhost:26257?sslmode=require' --insecure
```
I didn't realize until now that the order of the `--` flags mattered, but at least it does seem well-defined.

rafiss (approved):
LGTM! thanks

PR #278, opened 2023-07-22
docs: Inform about CI script

BuonOmo:
From [slack](https://cabesa.slack.com/archives/C01JBLP0CRX/p1689885055116999).

# Context

Sometimes the test suite is failing because of this kind of message:

```
TimestampsWithoutTransactionTest#test_do_not_write_timestamps_on_save_if_they_are_not_attributes = 0.18 s = E


Error:
TimestampsWithoutTransactionTest#test_do_not_write_timestamps_on_save_if_they_are_not_attributes:
RuntimeError: Wrapped undumpable exception for: ActiveRecord::StatementInvalid: PG::UndefinedTable: ERROR:  relation "p_timestamp_attribute_posts_s" does not exist
```

# Reason

It seems like the error comes from the `test/cases/migration/foreign_key_test.rb` file, and mostly from the fact that either the ensure block might not execute completely correctly.

Unfortunately this theory has some flaws: the test is passing, so there is no obvious failure. Hence this PR to ensure that this is a path we should continue investigating. And some documentation that would be helpful for any maintainer of this gem. (I could split this info in a MAINTAINER.md file if you prefer @rafiss)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,6 +36,8 @@ Ruby version.
 tests similarly to TeamCity. The database is destroyed between each
 test file.)
 
+Note that the actual CI test suit is ran with [this script](https://github.com/cockroachdb/cockroach/blob/master/pkg/cmd/roachtest/tests/activerecord.go). And test failures are tracked [cockroach#97283](https://github.com/cockroachdb/cockroach/issues/97283)
+
 Install rbenv with ruby-build on MacOS:
 
 ```bash
--- a/test/cases/migration/foreign_key_test.rb
+++ b/test/cases/migration/foreign_key_test.rb
@@ -284,22 +284,26 @@ module ActiveRecord
 
         def test_add_foreign_key_with_prefix
           ActiveRecord::Base.table_name_prefix = "p_"
+          puts "TEMP:1A: p_ set (#{__FILE__}:#{__LINE__ - 1})"
           migration = CreateSchoolsAndClassesMigration.new
           silence_stream($stdout) { migration.migrate(:up) }
           assert_equal 1, @connection.foreign_keys("p_classes").size
         ensure
           silence_stream($stdout) { migration.migrate(:down) }
           ActiveRecord::Base.table_name_prefix = nil
+          puts "TEMP:1B: p_ unset  (#{__FILE__}:#{__LINE__ - 1})"
         end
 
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -36,6 +36,8 @@ Ruby version.
 tests similarly to TeamCity. The database is destroyed between each
 test file.)
 
+Note that the actual CI test suit is ran with [this script](https://github.com/cockroachdb/cockroach/blob/master/pkg/cmd/roachtest/tests/activerecord.go). And test failures are tracked [cockroach#97283](https://github.com/cockroachdb/cockroach/issues/97283)
```
andyyang890, line 39: The CI test suite on the adapter repo is not run with that script, it's for running the nightly roachtest run of the activerecord test in the [cockroach](https://github.com/cockroachdb/cockroach/) repo.

I would also not hardcode the issue number because this is only the current issue tracking the master branch. Once the tests are fixed, it will be closed and a new issue will be opened for future test failures.

rafiss: That's right - the adapter CI uses this script actually: https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/002b0973e68211bbbe3f112f669b2c91a555ea4a/build/teamcity-test.sh#L71-L74

This is defined in the TeamCity build job: https://teamcity.cockroachdb.com/buildConfiguration/Cockroach_ScratchProjectPutTcExperimentsInHere_ActiveRecordTestSuite?mode=builds

As a future project, it would be simpler to execute the CI using Github actions (which wasn't available at the time when we first set this project up). This would make it simpler so we don't have to go to a different place to configure CI for this project. I made an issue: https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/279

As far the where the test failures are tracked: for our nightlies, this is the place we can link to: https://github.com/cockroachdb/cockroach/blob/8d10f15db8e5efaff57ec332c438211895c6c608/pkg/cmd/roachtest/tests/activerecord_blocklist.go#L36

The ideal state is that none of them should fail - either we skip them if we determine it cannot work with CRDB, or we fix the bug in the tests so it passes. I don't think we need to link to this test tracker here.

BuonOmo:
Closing in favor of #279. And once I'll see a reproduction of the bug mentionned by @andyyang890 I'll act on it!

PR #280, opened 2023-07-23
chore(ci): switch to github actions

### LINKED ISSUE #279: Move CI testing to Github actions
When this project was first set up, Github actions was not available. Right now the testing is executed in TeamCity: https://teamcity.cockroachdb.com/buildConfiguration/Cockroach_ScratchProjectPutTcExperimentsInHere_ActiveRecordTestSuite?mode=builds

It would be more maintainable if the testing were done with Github actions now. We have done this for other adapters, like this: https://github.com/cockroachdb/sqlalchemy-cockroachdb/blob/master/.github/workflows/ci.yml

BuonOmo:
When this project was first started, Github Actions were not an option. Now that they are, and that teamcity is less accessible (lesser used, needs connection, ci link is not up to date), we are switching to gh actions.

Fixes #279


Gemfile
```diff
@@ -14,7 +14,7 @@ module RailsTag
 
     def gemspec_requirement
       File
-        .foreach("activerecord-cockroachdb-adapter.gemspec", chomp: true)
+        .foreach(File.expand_path("activerecord-cockroachdb-adapter.gemspec", __dir__), chomp: true)
```
BuonOmo, line 17: so we can run bundler tasks from any subdirectory

test/cases/adapters/postgresql/active_schema_test.rb
```diff
@@ -64,8 +64,8 @@ def test_add_index
     end
 
     private
-      def method_missing(method_symbol, *arguments)
-        ActiveRecord::Base.connection.send(method_symbol, *arguments)
+      def method_missing(...)
+        ActiveRecord::Base.connection.send(...)
```
BuonOmo, line 68: fixes tests as ruby 3 stopped interpreting keyword args as normal args

test/cases/connection_adapters/schema_cache_test.rb
```diff
@@ -15,7 +15,7 @@ def setup
       # See test/excludes/ActiveRecord/ConnectionAdapters/SchemaCacheTest.rb
       def test_yaml_loads_5_1_dump
         body = File.open(schema_dump_path).read
-        cache = YAML.load(body)
+        cache = YAML.unsafe_load(body)
```
BuonOmo, line 18: fixes tests, psych got more strict over the years

test/cases/helper_cockroachdb.rb
```diff
@@ -1,5 +1,5 @@
-require 'bundler/setup'
-Bundler.require :development
+require 'bundler'
+Bundler.setup
```
BuonOmo, line 2: This is taken from https://github.com/rails/rails/blob/7f338b9082fe3488f681c171b93c371cfc951f22/load_paths.rb.

With the previous version we would load the `railties/lib/rails.rb` file from rails. This file defines `Rails.env`. And having this method defined keep a cached version of the env, failing tests that are trying to redefine the rails env...

BuonOmo:
As of now, here's the test summary result: `8015 runs, 25224 assertions, 12 failures, 263 errors, 29 skips`

test/excludes/ActiveRecord/ConnectionAdapters/PostgreSQLAdapter/QuotingTest.rb
```diff
@@ -0,0 +1,3 @@
+exclude :test_do_not_raise_when_int_is_not_wider_than_64bit, "replaced for correct quoting"
+exclude :test_do_not_raise_when_raise_int_wider_than_64bit_is_false, "replaced for correct quoting"
+exclude :test_raise_when_int_is_wider_than_64bit, "since integers are stringified there is not 64bit limit"
```
BuonOmo, line 3: I have to do some more documentation for this. Maybe @rafiss you know where to dig?

My current understanding of the context is that in ActiveRecord it seems that there is a 64bit cap for integer with the pg adapter. But since ints are stringified in crdb, there is no such thing for us.

BuonOmo:
@rafiss I've fixed loads of tests, some are still present but I think I'd rather focus on a new feature for the next week, and I'll see flaky tests coming at the same time! WDYT?

test/excludes/ActiveRecord/ConnectionAdapters/PostgreSQLAdapter/QuotingTest.rb
rafiss: i'm surprised - it shouldn't be possible to have an integer wider than 64 bits in CRDB.

test/cases/helper_cockroachdb.rb
rafiss: nice improvement

rafiss (approved):
i'm glad to see these improvements! and yes, let's merge this now, and we can make incremental progress on it

i'm not sure about the 64-bit test. should we understand it better before merging, or can we do that later too?

test/excludes/ActiveRecord/ConnectionAdapters/PostgreSQLAdapter/QuotingTest.rb
BuonOmo: Ok I'll dig further then

BuonOmo: So what I understood from this test: when quoting in pg, activerecord checks if the int is withing the 64 bit range with the `#check_int_in_range` method. The error message related to that is:

```ruby
        def check_int_in_range(value)
          if value.to_int > 9223372036854775807 || value.to_int < -9223372036854775808
            exception = <<~ERROR
              Provided value outside of the range of a signed 64bit integer.


              PostgreSQL will treat the column type in question as a numeric.
              This may result in a slow sequential scan due to a comparison
              being performed between an integer or bigint value and a numeric value.


              To allow for this potentially unwanted behavior, set
              ActiveRecord.raise_int_wider_than_64bit to false.
            ERROR
            raise IntegerOutOf64BitRange.new exception
          end
        end
```

https://github.com/rails/rails/blob/4e01a7e8753c30e22264e185082524a9ea5e0c90/activerecord/lib/active_record/connection_adapters/postgresql/quoting.rb#L28-L42

So PG would think of strings containing only numbers as int, and would be way slower if they exceed 64bit. @rafiss does this ring a bell to you? Is there a similar behaviour in cockroach?

I'll keep looking for documentation related to that in pg to be sure. Might edit this message when I find an answer.

The fix, if need be, is fairly easy.

rafiss: ah i understand now - so if it's larger than 64 bits it's treated as `decimal` (also called `numeric`) rather than `integer`

i think matching PG here makes sense. it's fine to make that change later too.

rafiss: thanks for the nice research on this!

BuonOmo: @rafiss sorry didn't see your message before editing. I think this needs our attention!

BuonOmo: After investigating some more, this adapter is not affected by the [mentioned vulnerability](https://github.com/advisories/GHSA-579w-22j4-4749). When quoting a number using this gem, it will be converted to a string first, taking advantage of cockroach's behavior of converting implicitly strings into numbers when needed. That conversion done on the server side is checking for bounds and raising if need be.

Here's an example of some code that could be abused in the pg adapter before the patch, but that raises in this adapter:

```ruby
# age is an int column with an index
Cat.where("age = ?", 2**63)
# => PG::InvalidTextRepresentation: ERROR:  could not parse "9223372036854775808" as type int: strconv.ParseInt: parsing "9223372036854775808": value out of range (ActiveRecord::StatementInvalid)
```

PR #283, opened 2023-08-13
fix: support multiple databases

BuonOmo:
This is following and replacing #271 now that I have repository access.


BuonOmo:
@rafiss I think further testing would require way too much time now and in tests. I'd stay as-is.

rafiss (approved):
lgtm!

PR #284, opened 2023-08-15
feat: Add AOST queries

### LINKED ISSUE #281: support for AOST queries
CockroachDB has support for AS OF SYSTEM TIME timestamp queries. This can be prefixed to any SELECT query to query using the systemtime and get a historical view of the data. More info here:
https://www.cockroachlabs.com/docs/v1.1/as-of-system-time

It would be good to have support for AOST queries in ActiveRecord. Some potential approaches using select examples from [here](https://api.rubyonrails.org/v7.0.6/classes/ActiveRecord/QueryMethods.html#method-i-select):

```ruby
Model.select(:field).aost(some_timestamp)
Client.where("orders_count = '2'").aost(some_timestamp)
```

I am not 100% certain on if adding .aost method will break things or would be an antipattern in ActiveRecord/Ruby. cc: @BuonOmo for guidance on that.

BuonOmo:
Add _As Of System Time_ support for models.

See https://www.cockroachlabs.com/docs/stable/as-of-system-time

Informs #281

### DIFF
--- a/test/cases/arel/nodes_test.rb
+++ b/test/cases/arel/nodes_test.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+# This file may be remove after next
+# rails release.
+# Whenever https://github.com/rails/rails/commit/8bd463c
+# is part of a rails version.
+
+require "cases/arel/helper"
+
+module Arel
+  module Nodes
+    class TestNodes < Arel::Test
+      def test_every_arel_nodes_have_hash_eql_eqeq_from_same_class
+        # #descendants code from activesupport
+        node_descendants = []
+        ObjectSpace.each_object(Arel::Nodes::Node.singleton_class) do |k|
+          next if k.respond_to?(:singleton_class?) && k.singleton_class?
+          node_descendants.unshift k unless k == self
+        end
+        node_descendants.delete(Arel::Nodes::Node)
+        node_descendants.delete(Arel::Nodes::NodeExpression)
+
+        default_hash_owner = Object.instance_method(:hash).owner
+
+        bad_node_descendants = node_descendants.reject do |subnode|
+          eqeq_owner = subnode.instance_method(:==).owner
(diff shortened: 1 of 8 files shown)


test/cases/arel/nodes_test.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
```
BuonOmo, line 1: This can be removed on next rails update as it was fixed upstream

rafiss: could a comment be added that mentions this?

rafiss (approved):
this looks good to me!

CRDB also supports using it like ` ... AS OF SYSTEM TIME '-10s'` (allowing any negative interval), as well as a few special functions for the parameter, described here: https://www.cockroachlabs.com/docs/stable/as-of-system-time#parameters

i think we can merge this PR, but could those other variants be supported too?

test/cases/arel/nodes_test.rb
BuonOmo: There is also one (see the excluded folder, it is mentionning the given PR). But I can do more for sure :)

BuonOmo:
@rafiss I saw those, and since activesupport is also providing such features (`10.seconds.ago` is valid ruby !) I was thinking of keeping it simple. Also for the reason that the feature for now only work out of transaction (see my comment in the issue #281) I don't think it is worth making it more complex for now. What do you think ?

As for the other functions, I completely overlooked them, lets keep the AOST issue open and discuss the extended scope over there if that's ok with you!

rafiss:
sounds great, thanks!

grantpaulson6:
When do you guys think this will be released?

Also, just wanted to bump the comment about special functions for the AOST query, would be a nice to have

BuonOmo:
@rafiss I don't see any particular reason not to release. WDYT?

PR #286, opened 2023-09-02
feat: support schema search path

BuonOmo:


test/excludes/SchemaTest.rb
```diff
@@ -1,38 +1,43 @@
...
-exclude :test_current_schema, "Custom schemas are not supported in CockroachDB. See https://github.com/cockroachdb/cockroach/issues/26443."
-exclude :test_prepared_statements_with_multiple_schemas, "Custom schemas are not supported in CockroachDB. See https://github.com/cockroachdb/cockroach/issues/26443."
-exclude :test_schema_exists?, "Custom schemas are not supported in CockroachDB. See https://github.com/cockroachdb/cockroach/issues/26443."
-exclude :test_reset_pk_sequence, "Custom schemas are not supported in CockroachDB. See https://github.com/cockroachdb/cockroach/issues/26443."
-exclude :test_set_pk_sequence, "Custom schemas are not supported in CockroachDB. See https://github.com/cockroachdb/cockroach/issues/26443."
+exclude :test_set_pk_sequence, "Skipping until we can triage further. See https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/48"
+exclude :test_pk_and_sequence_for_with_schema_specified, "Skipping until we can triage further. See https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/48"
```
BuonOmo, line 2: @rafiss these one are here because of the last bugfix you made in crdb (see #287)

BuonOmo:
One of the failing tests is new to this PR, but I'm not sure how to fix it.

We create a table using `id integer`, this is stored as `INT8` in crdb. but when we use the the method `column_definitions`, it is shown as `bigint`. This `bigint` value is found in this query:

```sql
SELECT a.attname, format_type(a.atttypid, a.atttypmod),
        pg_get_expr(d.adbin, d.adrelid), a.attnotnull, a.atttypid, a.atttypmod,
        c.collname, NULL AS comment,
        attgenerated,
        NULL as is_hidden
  FROM pg_attribute a
  LEFT JOIN pg_attrdef d ON a.attrelid = d.adrelid AND a.attnum = d.adnum
  LEFT JOIN pg_type t ON a.atttypid = t.oid
  LEFT JOIN pg_collation c ON a.attcollation = c.oid AND a.attcollation <> t.typcollation
  WHERE a.attrelid = 'thetable'::regclass
    AND a.attnum > 0 AND NOT a.attisdropped
  ORDER BY a.attnum
```

@rafiss do you have any idea if this is correct or expected ?

rafiss:
Yeah `bigint` is just an alias for `INT8`. Actually, it seems that Postgres also returns `bigint` for this case, so maybe something else is going wrong?

```
postgres=# create table thetable(a int8, b int4);
CREATE TABLE

postgres=# SELECT a.attname, format_type(a.atttypid, a.atttypmod),
        pg_get_expr(d.adbin, d.adrelid), a.attnotnull, a.atttypid, a.atttypmod,
        c.collname, NULL AS comment,
        attgenerated,
        NULL as is_hidden
  FROM pg_attribute a
  LEFT JOIN pg_attrdef d ON a.attrelid = d.adrelid AND a.attnum = d.adnum
  LEFT JOIN pg_type t ON a.atttypid = t.oid
  LEFT JOIN pg_collation c ON a.attcollation = c.oid AND a.attcollation <> t.typcollation
  WHERE a.attrelid = 'thetable'::regclass
    AND a.attnum > 0 AND NOT a.attisdropped
  ORDER BY a.attnum;
 attname | format_type | pg_get_expr | attnotnull | atttypid | atttypmod | collname | comment | attgenerated | is_hidden
---------+-------------+-------------+------------+----------+-----------+----------+---------+--------------+-----------
 a       | bigint      | NuLL        | f          |       20 |        -1 | NuLL     | NuLL    |              | NuLL
 b       | integer     | NuLL        | f          |       23 |        -1 | NuLL     | NuLL    |              | NuLL
(2 rows)
```

BuonOmo:
Ah so the problem seems to be at creation then: we say create with integer and we get a bigint. Thank you! I'll look at this further during the next days

BuonOmo:
@rafiss so looking again at the issue, I think there is a CRDB issue.

Those statements show me that id is an int8, which 

```sql
create table bar (id integer, num bigint);
show create table bar;
```

Shows both as the same `bigint`.

However doing the same with pg gives me integer and bigint.

rafiss:
Ah I see, yes there is a CRDB difference here. In CRDB, the default integer size is 8. It can be configured using the `default_int_size` session setting. (`SET default_int_size = 4`)

There are more docs here: https://www.cockroachlabs.com/docs/stable/int

BuonOmo:
@rafiss nice thank you ! I think I'll just set this in the database for tests to be ok and more in sync with the PG tests. I can also consider adding a dedicated test to show that this specific case of integer being the same as bigint is considered and not a bug.

BuonOmo:
The `default_int_size` issue is more problematic than expected. I found a solution for the tests which is to also change the `serial_normalization` value. Without that, serial and int could be represented with different limits, which messes with rails internal (I've seen that on a many2many join table).

For this PR, I think we're fine as this is out of scope anyway. But shouldn't I open an issue to discuss that and what do we want to do ? For now I see two options:
- tweak rails internal to use serial for its join tables, and maybe some other places, and play with the `default_int_size` setting in tests to be sure we have everything covered
- find a way to warn users about setting `default_int_size` without changing `serial_normalization`.

WDYT @rafiss ?

rafiss:
> shouldn't I open an issue to discuss that and what do we want to do ? For now I see two options:
>
> - tweak rails internal to use serial for its join tables, and maybe some other places, and play with the default_int_size setting in tests to be sure we have everything covered
> - find a way to warn users about setting default_int_size without changing serial_normalization.

I believe you encountered this problem: https://github.com/cockroachdb/cockroach/issues/26925#issuecomment-1255293916

But it's possible we don't show this warning in all the cases we should. For now, I agree: let's start by opening an issue in this repo, with a full example of the table schema that causes the confusing behavior.

rafiss (approved):
nice work on this!

PR #287, opened 2023-09-07
chore: unexclude tests fixed in CRDB

BuonOmo:
See https://github.com/cockroachdb/cockroach/pull/110144

@rafiss I can force push the commit again to re-run the test once crdb is updated!


BuonOmo:
Seems to me like we can now merge this one, can't we ? (note that you would have to change the branch protection rule pattern)

rafiss (approved):

PR #289, opened 2023-09-22
fix(schema): render schema name when dumping foreign keys

BuonOmo:
In PostgreSQL, `oid::regclass::text` renders the schema name if it is not `public`. In CockroachDB, this is not the case, so we need to append the schema name manually.

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
@@ -36,16 +36,24 @@ module ActiveRecord
         # Modified version of the postgresql foreign_keys method.
         # Replaces t2.oid::regclass::text with t2.relname since this is
         # more efficient in CockroachDB.
+        # Also, CockroachDB does not append the schema name in relname,
+        # so we append it manually.
         def foreign_keys(table_name)
           scope = quoted_scope(table_name)
           fk_info = exec_query(<<~SQL, "SCHEMA")
-            SELECT t2.relname AS to_table, a1.attname AS column, a2.attname AS primary_key, c.conname AS name, c.confupdtype AS on_update, c.confdeltype AS on_delete, c.convalidated AS valid
+            SELECT CASE
+              WHEN n2.nspname = current_schema()
+              THEN ''
+              ELSE n2.nspname || '.'
+            END || t2.relname AS to_table,
+            a1.attname AS column, a2.attname AS primary_key, c.conname AS name, c.confupdtype AS on_update, c.confdeltype AS on_delete, c.convalidated AS valid, c.condeferrable AS deferrable, c.condeferred AS deferred
             FROM pg_constraint c
             JOIN pg_class t1 ON c.conrelid = t1.oid
             JOIN pg_class t2 ON c.confrelid = t2.oid
             JOIN pg_attribute a1 ON a1.attnum = c.conkey[1] AND a1.attrelid = t1.oid
             JOIN pg_attribute a2 ON a2.attnum = c.confkey[1] AND a2.attrelid = t2.oid
             JOIN pg_namespace t3 ON c.connamespace = t3.oid
+            JOIN pg_namespace n2 ON t2.relnamespace = n2.oid
             WHERE c.contype = 'f'
               AND t1.relname = #{scope[:name]}
               AND t3.nspname = #{scope[:schema]}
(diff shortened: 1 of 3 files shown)


lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
```diff
@@ -36,10 +36,13 @@ def primary_key(table_name)
         # Modified version of the postgresql foreign_keys method.
         # Replaces t2.oid::regclass::text with t2.relname since this is
         # more efficient in CockroachDB.
+        # Also, CockroachDB does not append the schema name in relname,
+        # so we append it manually.
         def foreign_keys(table_name)
           scope = quoted_scope(table_name)
           fk_info = exec_query(<<~SQL, "SCHEMA")
-            SELECT t2.relname AS to_table, a1.attname AS column, a2.attname AS primary_key, c.conname AS name, c.confupdtype AS on_update, c.confdeltype AS on_delete, c.convalidated AS valid
+            SELECT CASE WHEN t2.relnamespace::regnamespace::text = 'public' THEN t2.relname ELSE t2.relnamespace::regnamespace::text || '.' END || t2.relname AS to_table,
```
BuonOmo, line 44: @rafiss if you have any other, more performant, idea, feel free to share!

Or maybe the change should be upstreamed to crdb? (e.g: `relname` should contain the schema if not public)

rafiss: hm, but i tested on postgres, and it does not seem to include the schema name if not pubic either. this is postgres behavior:
```
postgres=# create schema s;

postgres=# create table s.t(a int);
CREATE TABLE

postgres=# create table s.u(a int, b int  references t(a));
CREATE TABLE

postgres=# SELECT t2.relname AS to_table, a1.attname AS column, a2.attname AS primary_key, c.conname AS name, c.confupdtype AS on_update, c.confdeltype AS on_delete, c.convalidated AS valid
            FROM pg_constraint c
            JOIN pg_class t1 ON c.conrelid = t1.oid
            JOIN pg_class t2 ON c.confrelid = t2.oid
            JOIN pg_attribute a1 ON a1.attnum = c.conkey[1] AND a1.attrelid = t1.oid
            JOIN pg_attribute a2 ON a2.attnum = c.confkey[1] AND a2.attrelid = t2.oid
            JOIN pg_namespace t3 ON c.connamespace = t3.oid
            WHERE c.contype = 'f'
              AND t1.relname = 'u'
              AND t3.nspname = 's'
ORDER BY c.conname;
 to_table | column | primary_key |   name   | on_update | on_delete | valid
----------+--------+-------------+----------+-----------+-----------+-------
 t        | b      | a           | u_b_fkey | a         | a         | t
(1 row)
```

do you know why the schema name is needed for CRDB?

rafiss:
> In PostgreSQL, oid::regclass::text renders the schema name if it is not public. In CockroachDB, this is not the case, so we need to append the schema name manually.

This looks like a mistake. the behavior seems the same with CRDB and postgres:
```
# Postgres
postgres=# select 33391::regclass::text;
 text
-------
 33391
(1 row)

postgres=# select 33391::regnamespace::text;
 text
------
 s
(1 row)
```

```
# CockroachDB

root@localhost:26257/defaultdb> select 104::regclass::text;;
  text
--------
  104
(1 row)

root@localhost:26257/defaultdb> select 104::regnamespace::text;;
  text
--------
  s
(1 row)
```

lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
BuonOmo: In postgres we do not use `t2.relname` but `t2.oid::regclass::text`, which renders everything (tested in pg as well!)

BuonOmo:
Try this one:

```sql
create schema foo;
create table foo.bar();
select oid::regclass::text from pg_class where relname = 'bar';
```

| crdb | pg |
| ---- | -- |
| `bar`| `foo.bar` |

lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
BuonOmo: In the pg adapter they use `oid::regclass::text` rather than `relname` (see my comment below)

rafiss (approved):

BuonOmo:
@rafiss I thought that you mentioned that issue (my last comment) upstream. I cannot find the related issue though. Do you remember doing such a thing ?

PR #290, opened 2023-09-22
fix(test): fk always valid

BuonOmo:
This fixes the test `#test_all_foreign_keys_valid_having_foreign_keys_in_multiple_schemas`.

@rafiss this makes it that we do not need the query to validate referential integrity (see slack).

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb/referential_integrity.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/referential_integrity.rb
@@ -11,6 +11,14 @@ module ActiveRecord
   module ConnectionAdapters
     module CockroachDB
       module ReferentialIntegrity
+        # CockroachDB will raise a `PG::ForeignKeyViolation` when re-enabling
+        # referential integrity (e.g: adding a foreign key with invalid data
+        # raises).
+        # So foreign keys should always be valid for that matter.
+        def all_foreign_keys_valid?
+          true
+        end
+
         def disable_referential_integrity
           foreign_keys = tables.map { |table| foreign_keys(table) }.flatten
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+root = true
+
+[*.rb]
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+insert_final_newline = true
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,11 @@ module RailsTag
     def call
       req = gemspec_requirement
       "v" + all_activerecord_versions.find { req.satisfied_by?(_1) }.version
+    rescue => e
+      warn "Unable to determine Rails version. Using last used. Error: #{e.message}"
+      lockfile = File.expand_path("Gemfile.lock", __dir__)
+      File.foreach(lockfile, chomp: true).find { _1[/tag: (.*)$/] }
+      Regexp.last_match(1)
     end
 
     def gemspec_requirement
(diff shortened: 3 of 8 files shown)


lib/active_record/connection_adapters/cockroachdb/referential_integrity.rb
```diff
@@ -11,6 +11,14 @@ module ActiveRecord
   module ConnectionAdapters
     module CockroachDB
       module ReferentialIntegrity
+        # CockroachDB will raise a `PG::ForeignKeyViolation` when re-enabling
+        # referential integrity (e.g: adding a foreign key with invalid data
+        # raises).
+        # So foreign keys should always be valid for that matter.
+        def all_foreign_keys_valid?
```
rafiss, line 18: I noticed that this method may be deprecated: https://github.com/rails/rails/blob/4b8cdef21054aaf6d23a2d2f90418b58e4d0c8ff/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L654-L661

I wonder if it's better to avoid using it. It seems like maybe one other option is to override `check_all_foreign_keys_valid` instead. What do you think?

Or another thing might be the `config.active_record.verify_foreign_keys_for_fixtures` option, which looks it will disable these checks in the first place https://github.com/rails/rails/blob/4b8cdef21054aaf6d23a2d2f90418b58e4d0c8ff/guides/source/configuring.md?plain=1#L1286. It's referenced [here](https://github.com/rails/rails/blob/4b8cdef21054aaf6d23a2d2f90418b58e4d0c8ff/activerecord/lib/active_record/fixtures.rb#L677).


If we do keep this method, I'm not sure I understand this - does this mean that the adapter will always consider all foreign keys to be valid? Could it be changed so that `all_foreign_keys_valid?` only returns true if `disable_referential_integrity` was called?

BuonOmo: @rafiss nice find! It seems like shipping with next AR minor.

I'll keep this one for now, and when shipping next AR, update `check_all_foreign_keys_valid!` to be sure we don't check anything.

> If we do keep this method, I'm not sure I understand this - does this mean that the adapter will always consider all foreign keys to be valid? Could it be changed so that all_foreign_keys_valid? only returns true if disable_referential_integrity was called?

The idea is to use the same method as the abstract one, I could make it more obvious by using this bit of code:

```ruby
def self.included(base)
  # Getting rid of PG specific check as in CRDB there are no trigger,
  # and thus no such thing as an invalid foreign key.
  base.remove_method(:check_all_foreign_keys_valid!)
```

About the option, I think the documentation is correct: this is PG/sqlite specific!

rafiss: >  # Getting rid of PG specific check as in CRDB there are no trigger,
>  # and thus no such thing as an invalid foreign key.

That's not quite accurate - CRDB does have the concept of an invalid foreign key, it's just that they are not implemented using triggers. (I know you are on vacation now, but just leaving this note to keep track of this discussion.)

rafiss (approved):

lib/active_record/connection_adapters/cockroachdb/referential_integrity.rb
BuonOmo: @rafiss still available a bit :)

Feel free to change the comment. I extrapolated from what I saw: it is not possible using the same method `disable_referential_integrity` to generate the same incoherence in the DB.

rafiss: ah yes that does seem correct - CRDB doesn't have a way to "turn off" referential integrity checks other than just removing the foreign key constraint entirely. i think we can leave things as they are.

PR #291, opened 2023-09-23
fix(test): Ensure `$stdout` is always restored

BuonOmo:
This fixes the `IOError` we could encounter in tests.

Found using:

```ruby
def bugmsg(header, stdout)
  bugged = begin; print; rescue IOError; "💥 "; end
  STDOUT.puts "🤞#{header.rjust(17)}: stdout=#{stdout.inspect} " \
    "#{bugged}(at #{caller_locations(2, 1).first})"
end

trace_var :$stdout, (proc { bugmsg("$stdout=", _1) })

TracePoint.trace(:a_return) do |tp|
  next unless tp.method_id == :reopen && tp.self == $stdout
  bugmsg("$stdout#reopen", $stdout)
end
```

### DIFF
--- a/test/cases/schema_dumper_test.rb
+++ b/test/cases/schema_dumper_test.rb
@@ -83,7 +83,7 @@ module CockroachDB
         end
       ensure
         migration.migrate(:down)
-        # $stdout = original
+        $stdout = original
       end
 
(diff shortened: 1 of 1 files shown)


rafiss (approved):
very nice find!

PR #292, opened 2023-09-25
chore(test): add github summary

BuonOmo:
I've been working on this little gem on my spare time, I think it would be quite useful for us! 

Here's the result (the *__Test Summary__* part): https://github.com/cockroachdb/activerecord-cockroachdb-adapter/actions/runs/6301821435


BuonOmo:
Added on rails 7.1 update

PR #294, opened 2023-09-27
DRAFT: feat: support datetime precision

BuonOmo:
This fixes some failing tests AND add a new feature :)

The only thing that worries me is that default of precision 6, which makes some of the compatibility test nonsensical.

I've talked about it in https://github.com/cockroachdb/cockroach/issues/110787#issuecomment-1737754509

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Ongoing
 
+- Add support for datetime with given precision ([#294](https://github.com/cockroachdb/activerecord-cockroachdb-adapter/pull/294))
+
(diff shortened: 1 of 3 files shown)


BuonOmo:
I cannot officially mark it as draft again, but I have sever doubt about this PR: the default precision with pg is 0, and crdb 6. I don't know what that implies on every levels yet. And seeing the huge amount of failures, I don't think I'll have time to dig further soon.

BuonOmo:
Fixed by https://github.com/cockroachdb/activerecord-cockroachdb-adapter/commit/0c2678d79f2bcc66fc02d08ea4a847c63416a362

PR #295, opened 2023-10-03
workflows: only trigger on pull_request and add a job wrapper

rafiss:
Previously the action would be executed twice, once by a push event and
once by a pull_request event.

Add a wrapper job with a stable name, so it's easy to configure a
required check on it.


rafiss:
thanks for the review!

PR #300, opened 2024-01-07
feat: rails 7.1 support

BuonOmo:
Remove deprecated `configure_connection` method override. It is now unnecessary as CockroachDB supports `SET TIMEZONE <...>`. This fixes the unset `type_map` bug

Fix changes induced by https://github.com/rails/rails/commit/896d35910a3e5c7363533e63903902c4a7483f1e


lib/active_record/connection_adapters/cockroachdb/database_statements.rb
```diff
@@ -2,21 +2,6 @@ module ActiveRecord
   module ConnectionAdapters
     module CockroachDB
       module DatabaseStatements
-        # Since CockroachDB will run all transactions with serializable isolation,
-        # READ UNCOMMITTED, READ COMMITTED, and REPEATABLE READ are all aliases
-        # for SERIALIZABLE. This lets the adapter support all isolation levels,
-        # but READ UNCOMMITTED has been removed from this list because the
-        # ActiveRecord transaction isolation test fails for READ UNCOMMITTED.
-        # See https://www.cockroachlabs.com/docs/v19.2/transactions.html#isolation-levels
```
BuonOmo, line 10: I'm not sure since when it is as it is, but the content is the same as pg's.

lib/active_record/connection_adapters/cockroachdb/referential_integrity.rb
```diff
@@ -39,16 +39,12 @@ def disable_referential_integrity
...
-                else
-                  raise error
-                end
-              end
+              # Avoid having PG:DuplicateObject error if a test is ran in transaction.
+              # TODO: verify that there is no cache issue related to running this (e.g: fk
+              #   still in cache but not in db)
+              next if foreign_key_exists?(foreign_key.from_table, name: foreign_key.options[:name])
+
+              add_foreign_key(foreign_key.from_table, foreign_key.to_table, **foreign_key.options)
```
BuonOmo, line 47: @rafiss any idea on the state of https://github.com/cockroachdb/cockroach/issues/19444 ? This would be quite helpful. Otherwise, I think there is some overall work on this because some fk are not correctly deleted...

lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
```diff
@@ -61,15 +62,26 @@ def foreign_keys(table_name)
...
+              column: column,
               name: row["name"],
-              primary_key: row["primary_key"]
+              primary_key: primary_key
             }
-
             options[:on_delete] = extract_foreign_key_action(row["on_delete"])
             options[:on_update] = extract_foreign_key_action(row["on_update"])
-            options[:deferrable] = extract_foreign_key_deferrable(row["deferrable"], row["deferred"])
+            options[:deferrable] = extract_constraint_deferrable(row["deferrable"], row["deferred"])
```
BuonOmo, line 84: I'm not sure about how well `deferrable` is supported by CRDB but most tests covering it are ignored, if there is full CRDB support, we should open a related issue to track this

lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -617,6 +532,26 @@ def load_additional_types(oids = nil)
...
+          else
+            yield query + initializer.query_conditions_for_known_type_names
+            yield query + initializer.query_conditions_for_known_type_types
+            yield query + initializer.query_conditions_for_array_types
+          end
+        end
+
+        def configure_connection
+          super
+          internal_execute("SET intervalstyle_enabled = true", "SCHEMA")
```
BuonOmo, line 552: I'm not sure we need it, I blindly followed recommendation of a previous comment. This might not be necessary anymore (depending on how crdb's configuration is)

test/cases/adapters/postgresql/postgresql_adapter_test.rb
```diff
@@ -63,23 +63,34 @@ def test_using_follower_reads_connects_properly
...
+        with_example_table do
+          @connection.exec_query("INSERT INTO ex (number) VALUES (1), (1)")
+          error = assert_raises(ActiveRecord::RecordNotUnique) do
+            @connection.add_index(:ex, :number, unique: true, algorithm: :concurrently, name: :invalid_index)
           end
+          assert_match(/duplicate key value violates unique constraint/, error.message)
+          assert_equal @connection.pool, error.connection_pool
+
+          # TODO: It seems like in CRDB this tests won't create the index at all.
+          #   It is created in PG with `valid: false` set. Is this correct?
```
BuonOmo, line 85: So here there is a behaviour difference between CRDB and PG:

PG creates the index anyway, but invalid. CRDB doesn't create it. Should I dig into it to see if this is a CRDB bug, or something we want ?

test/cases/adapters/postgresql/virtual_column_test.rb
```diff
@@ -3,18 +3,37 @@
...
+          t.virtual :upper_name,  type: :string,  as: "UPPER(name)", stored: true
+          t.virtual :name_length, type: :integer, as: "LENGTH(name)", stored: true
+          t.virtual :name_octet_length, type: :integer, as: "OCTET_LENGTH(name)", stored: true
+          t.integer :column1
+          t.virtual :column2, type: :integer, as: "column1 + 1", stored: true
+        end
+        VirtualColumn.create(name: "Rails")
+      end
+
+      # TODO: is this test result acceptable?
```
BuonOmo, line 29: Here's the CRDB dump
```ruby
  create_table "virtual_columns", force: :cascade do |t|
    t.string "name"
    t.virtual "upper_name", type: :string, as: "upper(name)", stored: true
    t.virtual "name_length", type: :bigint, as: "length(name)", stored: true
    t.virtual "name_octet_length", type: :bigint, as: "octet_length(name)", stored: true
    t.bigint "column1"
    t.virtual "column2", type: :bigint, as: nil, stored: true
  end
```

And here's the original pg expectation:
```ruby
t.virtual "upper_name", type: :string, as: "upper((name)::text)", stored: true
t.virtual "name_length", type: :integer, as: "length((name)::text)", stored: true
t.virtual "name_octet_length", type: :integer, as: "octet_length((name)::text)", stored: true
t.virtual "column2", type: :integer, as: "(column1 + 1)", stored: true
```

Does the current dump make sense to you @rafiss ?

test/cases/migration_test.rb
```diff
@@ -101,76 +103,38 @@ def setup
...
+      file, line = ::BulkAlterTableMigrationsTest.instance_method(method_name).source_location
+      iter = File.foreach(file)
+      (line - 1).times { iter.next }
+      indent = iter.next[/\A\s*/]
+      content = +""
+      content << iter.next until iter.peek == indent + "end\n"
+      content['"PostgreSQLAdapter" =>'] = '"CockroachDBAdapter" =>'
+      eval(<<-RUBY, binding, __FILE__, __LINE__ + 1)
+        def #{method_name}
+          #{content}
```
BuonOmo, line 121: I think we should go more in that direction to avoid having micro changes in the tests making our test suite fail. WDYT?

test/support/rake_helpers.rb
```diff
@@ -25,13 +25,13 @@ def test_files
   end
 
   def all_test_files
-    activerecord_test_files = Dir.
-      glob("#{ARTest::CockroachDB.root_activerecord}/test/cases/**/*_test.rb").
-      reject { _1.match?(%r(/adapters/(?:mysql2|sqlite3)/)) }.
-      prepend(COCKROACHDB_TEST_HELPER)
+    activerecord_test_files =
```
BuonOmo, line 28: The file order did matter, I opened a related issue in rails (https://github.com/rails/rails/issues/50768)

BuonOmo:
@rafiss I think you can already start reviewing. The majority of failing tests were already present in previous versions, others are documented, the only ones bothering me are the context dependent ones that do not trigger every time. I think we can still go on with this version as those were also present sometimes in previous versions...

lib/active_record/connection_adapters/cockroachdb/referential_integrity.rb
rafiss: unfortunately we aren't able to prioritize work on https://github.com/cockroachdb/cockroach/issues/19444 right now.

lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
rafiss: CRDB doesn't support `deferrable` right now: https://github.com/cockroachdb/cockroach/issues/31632

lib/active_record/connection_adapters/cockroachdb_adapter.rb
rafiss: we shouldn't need this now; it was only needed on v22.1 and earlier: https://github.com/cockroachdb/cockroach/pull/82388

test/cases/adapters/postgresql/postgresql_adapter_test.rb
rafiss: this is an area where CRDB diverges from PG. see: https://github.com/cockroachdb/cockroach/issues/65929

test/cases/adapters/postgresql/virtual_column_test.rb
rafiss: this one does seem like a CRDB bug. are you able to track down where the `as` field gets populated?

test/cases/migration_test.rb
rafiss: i don't quite understand this change, could you explain more?

test/cases/adapters/postgresql/virtual_column_test.rb
BuonOmo: Will look for it for sure. What about the other changes ? The `::text` mainly, since the `bigint` I think is ok as we already discussed that the naming and defaults of numbers are a bit different from pg

test/cases/adapters/postgresql/postgresql_adapter_test.rb
BuonOmo: So I should modify the test to look for a constraint then ? And add a comment referencing this issue maybe

lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
BuonOmo: Perfect, that's what I understood!

lib/active_record/connection_adapters/cockroachdb/referential_integrity.rb
BuonOmo: https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/309

test/cases/migration_test.rb
BuonOmo: Yeah definitely sorry. Basically I'm advocating for a bit of meta-programmation in tests.

## now

We currently are copying failing tests, and the related private methods. And then changing the bits we want to change.

## the change

I'd say that if the change is trivial, we can take advantage of ruby's meta-programmation tools to just change the parts we want. This would be way more resilient to little changes in the tests.

## the benefit and cost

This would make following rails updates so much easier, and even between minors or patches there are these kind of changes. It would reduce my workload. But it would make these local versions way harder to understand, as one would have to rely on understanding the meta change. But I think with good comment and viewing the original method, this should be ok

## the plan

I would make these changes progressively : with newly introduced tests, or with the one that are currently ignored and that we should stop ignoring anyway. Once there are a few examples I could refactor the meta-prog tools used (such as search and replace in the original function) to make the changes clearer and the footprint smaller

test/cases/adapters/postgresql/virtual_column_test.rb
BuonOmo: The `as` population is found and fixed in the latest commit !

lib/active_record/connection_adapters/cockroachdb_adapter.rb
BuonOmo: Weirdly the `NOTICE` about its usage still shows up in tests. Do you know what could be the cause ? Or how to track which query triggered a `NOTICE` event ?

test/cases/adapters/postgresql/postgresql_adapter_test.rb
BuonOmo: FYI @rafiss I've looked at what is generated and I see neither and index nor a constraint

rafiss: ah i see; i read the test more closely now. yes, that's another difference in CRDB. if you try to add the unique index but it fails, then neither the index nor constraint will exist.

test/cases/migration_test.rb
rafiss: thanks for explaining. i think the approach sounds good! i'd definitely like to make your maintenance burden easier.

rafiss (approved):
nice work!

looks like we are at:
```
8808 runs, 29847 assertions, 7 failures, 9 errors, 66 skips
```

are we going to address those last few failures now, or handle them separately?

also, do you think we should squash this all when merging?

BuonOmo:
> are we going to address those last few failures now, or handle them separately?

Those failures were here before already. I definitely plan on handling them but I didn't want this to be in the scope of the PR

> also, do you think we should squash this all when merging?

Cleaning up now !

PR #311, opened 2024-01-27
fix(schema_statements): add unique_rowid to pk_and_sequence_for

BuonOmo:
Allow to get the primary key in the default case.

This fixes two of the four failing `test_pk_and_sequence_for*`. The two still failing are:

## `test_pk_and_sequence_for_returns_nil_if_no_pk`

```ruby
      def test_pk_and_sequence_for_returns_nil_if_no_pk
        with_example_table "id integer" do
          assert_nil @connection.pk_and_sequence_for("ex") # => `["rowid", nil]`
        end
      end
```

the example table creation query:

```sql
CREATE TABLE ex(id integer)
```

The second query of `#pk_and_sequence_for` returns `["rowid", "public", nil]`. And the `SHOW CREATE TABLE ex` confirms it:

```sql
CREATE TABLE public.ex (
	id INT8 NULL,
	rowid INT8 NOT VISIBLE NOT NULL DEFAULT unique_rowid(),
	CONSTRAINT ex_pkey PRIMARY KEY (rowid ASC)
)
```

## `test_pk_and_sequence_for_with_collision_pg_class_oid`

```sql
create table ex(id serial primary key);
create table ex2(id serial primary key);
DELETE FROM pg_depend WHERE objid = 'ex_id_seq'::regclass AND refobjid = 'ex'::regclass AND deptype = 'a';
```

Gives the error:

> user root does not have DELETE privilege on relation pg_depend

The bug is not related to the code, but just to the tests. @rafiss could we give a user this privilege ? I didn't see it skimming the doc... (`show grants` gives `f` for `is_grantable` on `pg_depend`...):

```sql
grant DELETE on pg_depend to root;
--> ERROR: invalid privilege type DELETE for virtual_table
```

---

On meta-testing. This took me longer than expected. I'm a bit unsure about the return on investment for simple tests, as manipulating AST can feel a bit tricky. I still pursued it because it seems that the gems (parser and unparser) are quite stable, and if we could use this not only in tests, that would be tremendous. We would need more guaranties though. And we'd need to verify the performances..

I was thinking about checking the hash of the method to be changed against a saved hash (which would be stored in an argument). If there is a change we warn. So:

```ruby
CopyCat.copy_methods(..., hash: "somehash") do ... end
# => WARN: the method XXX was updated, here's the new changed method: ...
#      If it is ok change the hash to "somenewhash"
```

I think this is the way to go, but there are definitely some tweaks to make it easier.

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
@@ -57,6 +57,7 @@ module ActiveRecord
 
         # OVERRIDE: Added `unique_rowid` to the last line of the second query.
         #   This is a CockroachDB-specific function used for primary keys.
+        #   Also make sure we don't consider `NOT VISIBLE` columns.
         #
         # Returns a table's primary key and belonging sequence.
         def pk_and_sequence_for(table) # :nodoc:
@@ -68,7 +69,11 @@ module ActiveRecord
                  pg_attribute  attr,
                  pg_depend     dep,
                  pg_constraint cons,
-                 pg_namespace  nsp
+                 pg_namespace  nsp,
+                 -- TODO: use the pg_catalog.pg_attribute(attishidden) column when
+                 --   it is added instead of joining on crdb_internal.
+                 --   See https://github.com/cockroachdb/cockroach/pull/126397
+                 crdb_internal.table_columns tc
             WHERE seq.oid           = dep.objid
               AND seq.relkind       = 'S'
               AND attr.attrelid     = dep.refobjid
@@ -76,6 +81,9 @@ module ActiveRecord
               AND attr.attrelid     = cons.conrelid
               AND attr.attnum       = cons.conkey[1]
               AND seq.relnamespace  = nsp.oid
+              AND attr.attrelid     = tc.descriptor_id
+              AND attr.attname      = tc.column_name
+              AND tc.hidden         = false
               AND cons.contype      = 'p'
               AND dep.classid       = 'pg_class'::regclass
               AND dep.refobjid      = #{quote(quote_table_name(table))}::regclass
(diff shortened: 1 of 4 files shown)


lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
```diff
@@ -32,6 +32,63 @@ def primary_key(table_name)
...
 
+        # OVERRIDE: Added `unique_rowid` to the last line of the second query.
+        #   This is a CockroachDB-specific function used for primary keys.
+        #
+        # Returns a table's primary key and belonging sequence.
+        def pk_and_sequence_for(table) # :nodoc:
+          # First try looking for a sequence with a dependency on the
+          # given table's primary key.
+          result = query(<<~SQL, "SCHEMA")[0]
+            SELECT attr.attname, nsp.nspname, seq.relname
```
BuonOmo, line 43: @rafiss does this make sense for CRDB ? All the tests I'm working with until now are giving a void result...

BuonOmo:
Somehow the test `test_pk_and_sequence_for_returns_nil_if_no_pk` works locally and not on CI...

lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
rafiss: yes, this looks correct

if you are getting no results, it's probably because CRDB defaults to not using sequences for SERIAL types. this can be changed with the serial_normalization setting: https://www.cockroachlabs.com/docs/stable/serial#modes-of-operation

rafiss (commented):
> Gives the error:
> ```
> user root does not have DELETE privilege on relation pg_depend
> ```
> The bug is not related to the code, but just to the tests. @rafiss could we give a user this privilege ? I didn't see it skimming the doc... (show grants gives f for is_grantable on pg_depend...):

Unfortunately, pg_depend is view-only in CRDB, unlike PG. In CRDB there is no way to "artificially break" the dependency that the sequence has on the table.

BuonOmo:
About `test_pk_and_sequence_for_returns_nil_if_no_pk`. The table creation returns a line:

```
rowid INT8 NOT VISIBLE NOT NULL DEFAULT unique_rowid()
```

This line is the one that is failing us...

lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
```diff
@@ -68,14 +68,19 @@ def pk_and_sequence_for(table) # :nodoc:
                  pg_attribute  attr,
                  pg_depend     dep,
                  pg_constraint cons,
-                 pg_namespace  nsp
+                 pg_namespace  nsp,
+                 -- TODO: use the pg_catalog.pg_attribute(attishidden) column when it's added instead of joining on crdb_internal.
```
rafiss, line 72: it could be useful to link to https://github.com/cockroachdb/cockroach/pull/126397 in the TODO, so we can easily see when the new column is in a CRDB release.

BuonOmo:
It should be good for merging now! I think I'll work on the transaction related issues next, and then we'll almost be done with those red tests.

rafiss (approved):
lgtm!

PR #316, opened 2024-02-14
fix(#schema_names): Do not show crdb_internal schema

### LINKED ISSUE #314: crdb_internal in schema.rb
Hello,

I try to update our application to rails 7 and after gem updates and migrations in the schema.rb is `create_schema "crdb_internal"` added.

With this addition does `rails db:setup` which is used in CI and development not work anymore. 
It does not work, because crdb_internal already exists.

How I may avoid this line after migration?

```
ActiveRecord::Schema[7.1].define(version: 2024_02_12_145951) do
  create_schema "crdb_internal"
```

activerecord-cockroachdb-adapter (7.1.0)
rails (7.1.3)

After `rails db:drop` I cant do `db:setup` because of:
```
PG::DuplicateSchema: ERROR:  schema "crdb_internal" already exists (PG::DuplicateSchema)
```

BuonOmo:
The `schema_names` method is used to list the schemas in the database. It should not list internal schemas.

Implementation detail: we use the less performant ruby implementation rather than rewritting the method for maintainance costs.

Fixes #314

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Ongoing
 
 - Support arbitrary max identifier length ([#317](https://github.com/cockroachdb/activerecord-cockroachdb-adapter/pull/317)).
+- Fix `#schema_names` not to return `crdb_internal` ([#316](https://github.com/cockroachdb/activerecord-cockroachdb-adapter/pull/316))
 
 ## 7.1.0 - 2024-01-30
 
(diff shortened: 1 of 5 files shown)


rafiss (approved):
nice find!

jayzalowitz:
This is preventing me from starting a new project on crdb rails. How can I help @BuonOmo

BuonOmo:
@jayzalowitz no need to help ! I'll ping @rafiss for merging :)

PR #318, opened 2024-02-17
fix: Enable precision support

### LINKED ISSUE #307: Timestamp precision
The default timestamp precision is different in PG and CRDB, and behaviours differs. This lead to a lot of errored tests. I'll reference those here to treat the issue all at once.

- [x] test_timestamps_sets_default_precision_on_create_table
- [x] test_datetime_doesnt_set_precision_on_change_column
- [x] test_datetime_sets_precision_6_on_change_column
- [x] test_datetime_sets_precision_6_on_create_table
- [x] test_datetime_sets_precision_6_on_change_table

BuonOmo:
It was not enabled (due to [a bug in crdb](https://github.com/cockroachdb/cockroach/pull/111400)).
This allows behaviour closer to the pg adapter, and fixes some
tests. We also enhance the copy_cat to allow replacing a method
of the current class. And remove dead code.

Fixes #307


.ruby-version
```diff
@@ -1 +1 @@
-3.2.1
+3.2.3
```
BuonOmo, line 1: I'd still be very in favor of removing this file altogether...

lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -155,20 +154,6 @@ def max_transaction_retries
         @max_transaction_retries ||= @config.fetch(:max_transaction_retries, 3)
       end
 
-      # CockroachDB 20.1 can run queries that work against PostgreSQL 10+.
-      def postgresql_version
-        100000
-      end
```
BuonOmo, line 161: This bit I'm not sure about, let's see what the tests are saying about it

BuonOmo:
A lot of tests where failing with 23.1, and are working with 23.2. I think we'll have to go back to a feature flag.

lib/active_record/connection_adapters/cockroachdb_adapter.rb
BuonOmo: Looked around, this is a bit of an issue because of some of the `support...?` method in the pg adapter are depending on this for checks, and the crdb version is way higher than pg, making every check true. But it is just a matter of copying those checks and verifying if they are crdb compliant. Overall, I think we're better as we are now

rafiss (approved):

PR #322, opened 2024-03-03
fix: do not flush cache unnecessarily

### LINKED ISSUE #320: The activerecord-cockroachdb-adapter v7.1.0 throws an error when `use_follower_reads_for_type_introspection` is set to `true`
Your ruby gem activerecord-cockroachdb-adapter had a recent release to support rails 7.1. When using it, and the setting `use_follower_reads_for_type_introspection` = true, the error: "PG::FeatureNotSupported: ERROR: inconsistent AS OF SYSTEM TIME timestamp; expected: 1708547400.494068015,0, got: 1708547390.502956628,0
HINT: try SET TRANSACTION AS OF SYSTEM TIME" is raised from this line: https://github.com/cockroachdb/activerecord-cockroachdb-adapter/blob/9506aa79785e7e945ae86bc6579ce7330e38635e/lib/active_record/connection_adapters/cockroachdb_adapter.rb#L592

No application queries can execute, the error is triggered before they can run.

When setting `use_follower_reads_for_type_introspection` = false, there isn't any error.

**More info:** 
The error doesn't occur locally, it occurs using our qa cluster. And it appears the method that triggers it is `create_or_find_by!`
I think the error doesn't occur locally because its not a multi-region db, and it has different table localities.

I confirmed the error behavior in another application of ours too.

Weird thing is, not all rails models trigger it when calling `create_or_find_by!`.  My guess is that a unique index or foreign key might have something to do with which models trigger it? Hard to discern a pattern.

One other note: if the error doesn't occur first (isn't the first query run) it doesn't occur at all. Meaning, if I run some other query, then the offending `create_or_find_by!`, it doesn't throw an error.

Conversely, when the error does occur, all other subsequent queries will trigger an error. This error is different:

```
/home/code/vendor/bundle/ruby/3.2.0/gems/pg-1.5.5/lib/pg/result.rb:17:in `type_map=': wrong argument type nil (expected PG::TypeMap) (TypeError)

self.type_map = type_map
```

BuonOmo:
Follows rails changes introduced in
[rails/rails#48069][rails_pr].
The `exec_query` now flushes the schema cache.
As we now our queries do not alter this cache,
we'd rather not flush.

Might fix #320

[rails_pr]: https://github.com/rails/rails/pull/48069


BuonOmo:
@grantpaulson6 This might fix your issue. Or might be the reason of your error anyway (if some of your custom code in `#configure_connection` contains an `exec_query` or `execute`. Would you be able to try this branch on your cluster ?

rafiss (approved):

PR #323, opened 2024-03-29
Debug SQlite connection in tests

BuonOmo:


BuonOmo:
We can use the `fake` adapter instead.

PR #324, opened 2024-04-09
fix: Refactor new connection

### LINKED ISSUE #320: The activerecord-cockroachdb-adapter v7.1.0 throws an error when `use_follower_reads_for_type_introspection` is set to `true`

BuonOmo:
The old method was calling `AbstractAdapter#initialize` with a connection object, rather than a hash. This was then setting `@unconfigured_connection`, which in turns forced to re-run `#configure_connection` when a first query was made. This is usually OK (although unnecessary) except if `use_follower_reads_for_type_introspection` is used, since it sets AOST for a single query, which fails within a transaction.

Fixes #320

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -55,6 +55,7 @@ group :development, :test do
   gem "debug"
   gem "minitest-excludes", "~> 2.0.1"
   gem "minitest-github_action_reporter", github: "BuonOmo/minitest-github_action_reporter", require: "minitest/github_action_reporter_plugin"
+  gem "ostruct", "~> 0.6"
 
   # Gems used for tests meta-programming.
   gem "parser"
--- a/lib/active_record/connection_adapters/cockroachdb_adapter.rb
+++ b/lib/active_record/connection_adapters/cockroachdb_adapter.rb
@@ -32,35 +32,15 @@ require_relative "../relation/query_methods_ext"
 ActiveRecord::ConnectionAdapters::CockroachDB.initial_setup
 
 module ActiveRecord
+  # TODO: once in rails 7.2, remove this and replace with a `#register` call.
+  # See: https://github.com/rails/rails/commit/22a26d7f74ea8f0d5f7c4169531ae38441cfd5e5#diff-2468c670eb10c24bd2823e42708489a336d6f21c6efc7e3c4a574166fa77bb22
   module ConnectionHandling
+    def cockroachdb_adapter_class
+      ConnectionAdapters::CockroachDBAdapter
+    end
+
     def cockroachdb_connection(config)
-      # This is copied from the PostgreSQL adapter.
-      conn_params = config.symbolize_keys.compact
-
-      # Map ActiveRecords param names to PGs.
-      conn_params[:user] = conn_params.delete(:username) if conn_params[:username]
-      conn_params[:dbname] = conn_params.delete(:database) if conn_params[:database]
-
-      # Forward only valid config params to PG::Connection.connect.
-      valid_conn_param_keys = PG::Connection.conndefaults_hash.keys + [:requiressl]
-      conn_params.slice!(*valid_conn_param_keys)
-
-      ConnectionAdapters::CockroachDBAdapter.new(
-        ConnectionAdapters::CockroachDBAdapter.new_client(conn_params),
-        logger,
-        conn_params,
-        config
-      )
-    # This rescue flow appears in new_client, but it is needed here as well
-    # since Cockroach will sometimes not raise until a query is made.
-    rescue ActiveRecord::StatementInvalid => error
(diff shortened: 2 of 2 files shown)


dikshant:
This is great! Thanks for fixing!

Gemfile
```diff
@@ -55,6 +55,7 @@ group :development, :test do
   gem "debug"
   gem "minitest-excludes", "~> 2.0.1"
   gem "minitest-github_action_reporter", github: "BuonOmo/minitest-github_action_reporter", require: "minitest/github_action_reporter_plugin"
+  gem "ostruct", "~> 0.6"
```
grantpaulson6, line 58: whats the reason for adding this?

BuonOmo: @grantpaulson6 ruby 3.5 stops shipping ostruct, it sends a ruby warning. And ruby warnings are errors in the rails test suite. See https://github.com/cockroachdb/activerecord-cockroachdb-adapter/actions/runs/8621739468/job/23631300792

lib/active_record/connection_adapters/cockroachdb_adapter.rb
```diff
@@ -255,6 +235,21 @@ def self.database_exists?(config)
...
+      # since Cockroach will sometimes not raise until a query is made.
+      rescue ActiveRecord::StatementInvalid => error
+        no_db_err_check1 = @connection_parameters && @connection_parameters[:dbname] && error.cause.message.include?(@connection_parameters[:dbname])
+        no_db_err_check2 = @connection_parameters && @connection_parameters[:dbname] && error.cause.message.include?("pg_type")
+        if no_db_err_check1 || no_db_err_check2
+          raise ActiveRecord::NoDatabaseError
+        else
+          raise ActiveRecord::ConnectionNotEstablished, error.message
+        end
+      end
```
BuonOmo, line 251: This is also outdated compared to the rails db, and not tested in our database, and I don't know if _since Cockroach will sometimes not raise until a query is made._ is still valid and what if refers to. Do you have an idea of how to reproduce this behaviour @rafiss ?

grantpaulson6:
@BuonOmo when do you think this change will be released?

BuonOmo:
I don't know, it will depend on @rafiss' review. Besides that I think there's no issue

grantpaulson6:
just bumping this @rafiss @BuonOmo would love to see it sooner rather than later

lib/active_record/connection_adapters/cockroachdb_adapter.rb
rafiss: I am not sure, but a guess I have is that it refers to the fact that CRDB allows you to connect even if the connection string refers to a database name that does not exist. But when AR tries using that database name to do something, CRDB will return an error.

rafiss (approved):

grantpaulson6:
@rafiss or @BuonOmo would you release this? We just had it happen in production and caused errors. It hosed a rails worker and we had to kill it

BuonOmo:
I still don't have the permissions to do it, I'll ping @rafiss

PR #333, opened 2024-07-05
Fix various tests

BuonOmo:


rafiss:
This seems like an expected difference, due to the differences in how CRDB vs PG has decided to implement SERIALIZABLE. The short explanation is that in CRDB SERIALIZABLE, reads block on in-progress writes for as long as those writes are in progress. However, PG does not have this "read block on write" behavior, and so rather than allowing the left-hand-side to execute, it must instead abort that transaction. Both are valid ways to implement SERIALIZABLE.

So for this test case, maybe we should just skip it.

BuonOmo:
@rafiss there are many test cases impacted, I think I'll rewrite them (or at least some of them) to be sure we keep coverage! Thank you for that explanation :)

rafiss:
No problem!

Are these tests explicitly using SERIALIZABLE even with Postgres? If it's just using the default isolation level, another option is to configure CRDB to use READ COMMITTED isolation by default.

BuonOmo:
> Are these tests explicitly using SERIALIZABLE even with Postgres?

Yes they are!

I'm also wondering when rewriting the tests : how can I generate a serialization failure in CRDB to have correct tests? I'm playing around now to find it, but if you have a clear example it might help me!

BuonOmo:
Here's the curated sql repro for the test `test_deadlock_raises_Deadlocked_inside_nested_SavepointTransaction`:

```sql
DROP TABLE IF EXISTS "samples";
DROP TABLE IF EXISTS "bits";

CREATE TABLE "samples" ("id" text primary key, "value" integer);
CREATE TABLE "bits" ("id" serial primary key, "value" integer);

INSERT INTO "samples" ("id", "value") VALUES ('s1', 1);
INSERT INTO "samples" ("id", "value") VALUES ('s2', 2);

-- thread left
SELECT "bits".* FROM "bits" LIMIT 1; -- attempts 0
SELECT "samples".* FROM "samples" WHERE "samples"."id" = 's2' LIMIT 1 FOR UPDATE; -- attempts 0
SAVEPOINT active_record_1; -- attempts 0


-- thread right
BEGIN; -- attempts 0
SELECT "bits".* FROM "bits" LIMIT 1; -- attempts 0
SELECT "samples".* FROM "samples" WHERE "samples"."id" = 's1' LIMIT 1 FOR UPDATE; -- attempts 0
SAVEPOINT active_record_1; -- attempts 0
UPDATE "samples" SET "value" = 1 WHERE "samples"."id" = 's2'; -- attempts 0


-- thread left
UPDATE "samples" SET "value" = 2 WHERE "samples"."id" = 's1'; -- attempts 0


-- thread right
RELEASE SAVEPOINT active_record_1; -- attempts 0
COMMIT; -- attempts 0


-- thread left
ROLLBACK TO SAVEPOINT active_record_1; -- attempts 0
SELECT "samples".* FROM "samples" WHERE "samples"."id" = 's2' LIMIT 1 FOR UPDATE; -- attempts 1
SAVEPOINT active_record_1; -- attempts 1
SELECT "samples".* FROM "samples" WHERE "samples"."id" = 's2' LIMIT 1 FOR UPDATE; -- attempts 2
SAVEPOINT active_record_1; -- attempts 2
SELECT "samples".* FROM "samples" WHERE "samples"."id" = 's2' LIMIT 1 FOR UPDATE; -- attempts 3
SAVEPOINT active_record_1; -- attempts 3
ROLLBACK; -- attempts 3
BEGIN; -- attempts 1
SELECT "bits".* FROM "bits" LIMIT 1; -- attempts 1
SELECT "samples".* FROM "samples" WHERE "samples"."id" = 's2' LIMIT 1 FOR UPDATE; -- attempts 0
SAVEPOINT active_record_1; -- attempts 0
ROLLBACK TO SAVEPOINT active_record_1; -- attempts 0
ROLLBACK; -- attempts 0
```

rafiss:
Nice investigation! So if disabling the automatic retry logic causes the proper error as expected, then I think it's fair for us to skip this test in the CRDB test suite. 

If there's a way to tell the test to turn off the auto-retry behavior, then we can keep the test. But I don't think we should spend time on adding that functionality.

BuonOmo:
Skipping for now. I'll see what's up with 7.2

Ready to merge when you are @rafiss !

PR #345, opened 2024-10-04
fix: handle indexes with a hidden compound

### LINKED ISSUE #344: Rails migration fails after upgrading from 24.1 to 24.2 - no indexes found & no support for composite primary keys
After upgrading from 24.1 to 24.2, there is a `no indexes found` error and a `no support for composite primary keys` when running rails migration. This seems to be happening exclusively on Regional by Row tables with PKs and indexes that include the hidden `crdb_region` column 

It looks like the culprit could be that the `crdb_region` column is now included in `pg_index.indkey`

**Issue 1**
```
== migrating ============================
-- remove_index(:redacted_table_name, "redacted_table_name_id", {:algorithm=>:concurrently})
rails aborted!
StandardError: An error has occurred, all later migrations canceled: (StandardError)

No indexes found on redacted_table_name with the options provided.

Caused by:
ArgumentError: No indexes found on redacted_table_name with the options provided. (ArgumentError)
            raise ArgumentError, "No indexes found on #{table_name} with the options provided."
Tasks: TOP => db:migrate
```

**Issue 2**
```
WARNING: Active Record does not support composite primary key.
redacted_table_name has composite primary key. Composite primary key is ignored.
```

BuonOmo:
This fixes regional related parts, hence fixes #344

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
+++ b/lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
@@ -55,6 +55,37 @@ module ActiveRecord
           end
         end
 
+        def primary_keys(table_name)
+          return super unless database_version >= 24_02_02
+
+          query_values(<<~SQL, "SCHEMA")
+            SELECT a.attname
+              FROM (
+                     SELECT indrelid, indkey, generate_subscripts(indkey, 1) idx
+                       FROM pg_index
+                      WHERE indrelid = #{quote(quote_table_name(table_name))}::regclass
+                        AND indisprimary
+                   ) i
+              JOIN pg_attribute a
+                ON a.attrelid = i.indrelid
+               AND a.attnum = i.indkey[i.idx]
+               AND NOT a.attishidden
+             ORDER BY i.idx
+          SQL
+        end
+
+        def column_names_from_column_numbers(table_oid, column_numbers)
+          return super unless database_version >= 24_02_02
+
+          Hash[query(<<~SQL, "SCHEMA")].values_at(*column_numbers).compact
+            SELECT a.attnum, a.attname
+            FROM pg_attribute a
+            WHERE a.attrelid = #{table_oid}
+            AND a.attnum IN (#{column_numbers.join(", ")})
+            AND NOT a.attishidden
+          SQL
+        end
+
         # OVERRIDE: CockroachDB does not support deferrable constraints.
         #   See: https://go.crdb.dev/issue-v/31632/v23.1
         def foreign_key_options(from_table, to_table, options)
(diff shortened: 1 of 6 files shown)


lib/active_record/connection_adapters/cockroachdb/schema_statements.rb
```diff
@@ -55,6 +55,37 @@ def primary_key(table_name)
...
+        def column_names_from_column_numbers(table_oid, column_numbers)
+          return super unless database_version >= 24_02_02
+
+          Hash[query(<<~SQL, "SCHEMA")].values_at(*column_numbers).compact
+            SELECT a.attnum, a.attname
+            FROM pg_attribute a
+            WHERE a.attrelid = #{table_oid}
+            AND a.attnum IN (#{column_numbers.join(", ")})
+            AND NOT a.attishidden
+          SQL
```
mmalek-sa, line 86: This is not specific to CRDB >= 24.0.2. On 24.1.2 unique constraints on a region by row table will combine the `crdb_region` and column on the dumped schema. Since `attishidden` is not available in CRDB 24.1.2 this seems to do the trick:

```ruby
def column_names_from_column_numbers(table_oid, column_numbers)
  Hash[query(<<~SQL, "SCHEMA")].values_at(*column_numbers).compact
    SELECT a.attnum, a.attname
    FROM pg_attribute a
    WHERE a.attrelid = #{table_oid}
    AND a.attnum IN (#{column_numbers.join(", ")})
    AND a.attisdropped <> 't'
    AND a.attname <> 'crdb_region'
  SQL
end
```

Without `attisdropped <> t`, I see weird `....pg.dropped.18...` strings in the schema dump. All this is happening after upgrade to `Rails 7.1.x` and Cockroach AR Adapter `7.1.x`. I can't reproduce it on Rails `7.0.x`.

When it comes to `unique_constraints`, I also see a problem with indexes with expressions. For example:

```ruby
    t.index "lower(name::STRING) ASC", name: "index_table_name_redacted_on_name_unique", unique: true
```

is becoming:
```ruby
    t.unique_constraint [], name: "index_table_name_redacted_on_name_unique"
```

mmalek-sa: I created an issue with a workaround for `unique_constraint` [here](https://github.com/cockroachdb/activerecord-cockroachdb-adapter/issues/347).

BuonOmo: @mmalek-sa I'm not targetting `24.0.2` but `24.2.2`, which is the version where the related issue started raising. We do not need this patch before (hence the conditionnal!)

rafiss (approved):
nice, thanks for identifying the fix!

PR #349, opened 2024-10-16
fix: Avoid dumping unique_constraint when attisdropped

### LINKED ISSUE #347: Empty or incorrect unique_constraint generation
There seems to be a fundamental difference between PostgreSQL implementation of `pg_constraint` vs CockroachDB's implementation. When ActiveRecord queries this table [here](https://github.com/rails/rails/blob/6f57590388ca38ed2b83bc1207a8be13a9ba2aef/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb#L654-L662), Postgres returns an empty array while CockroachDB returns list of all indexes with a unique constraint.

This causes ActiveRecord to swap `t.index` definitions with `t.unique_constraint` definitions [here](https://github.com/rails/rails/blob/6f57590388ca38ed2b83bc1207a8be13a9ba2aef/activerecord/lib/active_record/schema_dumper.rb#L226-L230). The problem with this is that Postgres Adapter doesn't seem to support conditions and functions for constraints properly. This results in these index definitions to change to a wrong definition when using CockroachDB:

```ruby
t.index "lower(name::STRING) ASC", name: "index_redacted_table_name_on_name_unique", unique: true
```

becomes:
```ruby
t.unique_constraint [], name: "index_redacted_table_name_on_name_unique"
```

and 

```ruby
t.index ["username"], name: "index_users_on_username_unique", , unique: true, where: "(deleted is false)"
```

becomes

```ruby
t.unique_constraint ["username"], name: "index_users_on_username_unique"
```

I'm able to fix this by monkey patching `SchemaStatements#unique_constraints` method to return an empty array as PostgreSQL does:

```ruby
def unique_constraints(_table_name)
  []
end
```

This is happening in Rails >= `7.1`.

BuonOmo:
Fixes #347


mmalek-sa (commented):
This still doesn't handle the indexes with where clauses, for example:

```ruby
ActiveRecord::Schema.define do
  create_table :payments, force: true do |t|
    t.text 'name'
    t.timestamp 'deleted_at'
    t.index 'lower(name::STRING) ASC', name: 'lower_name_index_on_payments', unique: true
    t.index ['name'], name: 'index_payments_on_name', unique: true, where: '(deleted_at IS NULL)'
  end
end
```

Will result in a schema dump that lacks `deleted_at IS NULL`:

```ruby
ActiveRecord::Schema[7.2].define(version: 0) do
  create_table "payments", force: :cascade do |t|
    t.text "name"
    t.datetime "deleted_at", precision: nil
    t.index "lower(name) ASC", name: "lower_name_index_on_payments", unique: true
    t.unique_constraint ["name"], name: "index_payments_on_name"
  end
end
```

BuonOmo:
> Columns with the UNIQUE constraint automatically have an [index](https://www.cockroachlabs.com/docs/v24.2/indexes) created with the name <table name>_<columns>_key. To avoid having two identical indexes, you should not create indexes that exactly match the UNIQUE constraint's columns and order.

(https://www.cockroachlabs.com/docs/stable/unique)

I think we can just consider every `unique_constraint` as an `index` then. 

The UNIQUE constraint depends on the automatically created index, so dropping the index also drops the UNIQUE constraint. Conversely, [dropping the UNIQUE constraint](https://www.cockroachlabs.com/docs/v24.2/alter-table#drop-constraint) also drops the automatically created index.

rafiss (approved):
thanks for this fix!

PR #356, opened 2024-12-10
bump activerecord to 8.0.0

BuonOmo:


rafiss (approved):
nice work! thanks for identifying the fixes and exclusions that were needed.

PR #361, opened 2025-01-10
Speedup tests

BuonOmo:
Following the lead of https://github.com/rails/rails/pull/47499 we ensure tables are always unlogged, and we dump only necessary parts of the schema. This should gain ~7mn


test/cases/create_unlogged_tables_test.rb
```diff
@@ -24,24 +23,18 @@ class Thing < ActiveRecord::Base
...
-
-      teardown do
-        @connection.drop_table TABLE_NAME, if_exists: true
-        ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.create_unlogged_tables = false
-      end
-
       # Cockroachdb ignores the UNLOGGED specifier.
       # https://github.com/cockroachdb/cockroach/issues/56827
       def test_unlogged_in_test_environment_when_unlogged_setting_enabled
+        @original_create_unlogged_tables = ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.create_unlogged_tables
```
rafiss, line 29: unfortunately, UNLOGGED does not have any effect in CRDB, but the rest of this change does seem helpful. thanks!

https://www.cockroachlabs.com/docs/stable/create-table#parameters

rafiss (approved):

test/cases/create_unlogged_tables_test.rb
BuonOmo: Ok so maybe I'll just remove unlogged related tests and configurations!

PR #362, opened 2025-01-22
refactor(fixtures): Avoid disabling referential integrity

BuonOmo:
Avoid 87% of the `#disable_referential_integrity` calls made
in fixture setup by just trying first without disabling and
just disabling in case of failure. Also batch foreign keys
removals and addition (for clarity only).


rafiss (approved):
nice idea! it's great to see the tests this fast now.

PR #372, opened 2025-03-26
feat: add support for CRDB v25.1

BuonOmo:


BuonOmo:
Note that this also tests rails 8.0.2, which we seem to support well :)

rafiss (approved):
thank you for this update!

PR #373, opened 2025-03-26
chore(ci): add parallel testing

BuonOmo:


BuonOmo:
Definitely not conclusive as-is.

PR #377, opened 2025-07-03
fix: use internal_execute instead of execute_and_clear

Maimer:
it was removed in https://github.com/rails/rails/commit/fd24e5bfc9540fc00764a59ddf39a993bbd63ba2

### DIFF
--- a/lib/active_record/connection_adapters/cockroachdb_adapter.rb
+++ b/lib/active_record/connection_adapters/cockroachdb_adapter.rb
@@ -492,9 +492,8 @@ module ActiveRecord
           if @config[:use_follower_reads_for_type_introspection]
             initializer = OID::TypeMapInitializer.new(type_map)
             load_types_queries_with_aost(initializer, oids) do |query|
-              execute_and_clear(query, "SCHEMA", [], allow_retry: true, materialize_transactions: false) do |records|
-                initializer.run(records)
-              end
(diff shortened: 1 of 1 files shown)


rafiss:
Thanks for submitting this! Could you add more information in the PR description that describes what functionality was broken? i.e., provide an example code snippet that was having issues or provide the error that you encountered before making this patch? This will help us with documenting the bugfix in release notes.

Maimer:
@rafiss I think the tests failed due to some flakiness. Any chance you can rerun it and let me know if there is anything else that needs to be done here?

BuonOmo:
I've rerun the tests. But by looking at them, the fairness_by_group test has already failed in other branches so I would merge this PR confidentely.

Maimer:
> We'll fix that later no big deal, thank you @Maimer!

Are you or @rafiss able to handle the release? This is blocking for me currently, so it would great if we could get that new version released. 🙏

Maimer:
> I do not have sufficient permissions. In the meantime you could refer to the commit in your gemfile (`gem "activerecord-cockroachdb-adapter", git: "git@github.com:cockroachdb/activerecord-cockroachdb-adapter", ref: "f01c835"`)

Thanks, I'll have to wait for @rafiss to publish. Using a git referenced gem doesn't allow me to verify the checksum of the gem itself, so while it's useful for testing / development, it's not the best practice for production.

dikshant:
@rafiss could you help cut out a release?

PR #378, opened 2025-07-10
fix(test): propagate setup and teardown

BuonOmo:
Fixes the undefined class variable error in tests.


BuonOmo:
Closed in favor of #379

PR #395, opened 2025-11-25
feat(ci): follow rails more closely

BuonOmo:
- Add a `RAILS_TAG` environment variable to specify the Rails reference to test against.
- Update CI workflows to dynamically determine supported CockroachDB versions from the official CockroachDB docs.
- Weekly run CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,8 @@ on:
   # Triggers the workflow on pull request events.
   pull_request:
     types: [opened, reopened, synchronize]
-
+  schedule:
+    - cron: 12 13 * * 1 # Weekly
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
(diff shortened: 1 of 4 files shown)


BuonOmo:
Unfortunately this PR has to be stale for a little bit, the time for us to decide where to go this that. Currently, we're testing edge version and others, but some breaking changes in edge cannot be backported, so we have to have a way to distinguish between various versions. This may be done by having workflows against our `stable` branches rather than a matrix for instance, and contributions splitted between branches. But this is a long-running topic, no emergencies.

The benefits would be:
- clear view of which branch is failing and why
- availability for people using rails edge to use this adapter
- easier to backport changes
- way better time to publish a new release after rails does (almost instant vs a few weeks).

@rafiss if you have any thoughts or would like to ping anyone about that?

@dikshant I believe you do not work on that subject anymore ?

rytaft:
@BuonOmo This sounds like a good plan. Please let me know what you need from us to move forward here (feel free to ping me on slack if this discussion should be taken offline).

